Issue #1, opened 2021-05-08
Duplicate Line


itdependsnetworks:
### Environment
* Python version:  3.7
* netutils version:  0.1.1

### Expected Behavior
A duplicate line to raise an error as such

### Observed Behavior

A duplicate line, leads to unintended "grabbing of config"

### Steps to Reproduce
 incoming PR to show issue.

jeffkala:
closed via #128

PR #2, opened 2021-05-08
add dupline test, currently failing

itdependsnetworks:
This test fails, and demonstrates the issue identified in #1


jeffkala:
duplicate line was discussed and there hasn't been a valid configuration use case so far to support duplicate lines at the `root` level of a configuration.  It should be documented.

itdependsnetworks:
Addressed in #128

PR #8, opened 2021-05-14
Adds docs for readthedocs

qduk:
This now dynamically pulls from the README for the respective section. Was also able to fix the decorated functions and use invoke instead of make. Now, in the `docs` folder you can you just run `invoke html` and it'll build your html files. 

Edit:
To create files locally.
1. Install requirements under docs folder `pip install -r docs/requirements.txt
2. Run `invoke html` in project root. 
3. File will be generated in `docs/build`

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,0 +1,4 @@
+m2r2==0.2.7
+Sphinx
+toml
+sphinx-rtd-theme
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,0 +1,93 @@
+# Configuration file for the Sphinx documentation builder.
+#
+# This file only contains a selection of the most common options. For a full
+# list see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+# pylint: disable=W,C
+import os
+import sys
+import toml
+
+sys.path.insert(0, os.path.abspath("../.."))
+toml_dict = toml.load("../../pyproject.toml")
+
+
+# -- Project information -----------------------------------------------------
+
+project = toml_dict["tool"]["poetry"]["name"]
+copyright = f"{toml_dict['tool']['poetry']['authors'][0]}, 2021"
+author = ",".join(toml_dict["tool"]["poetry"]["authors"])
+
+# The full version, including alpha/beta/rc tags
+release = toml_dict["tool"]["poetry"]["version"]
+
+
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = ["sphinx.ext.autodoc", "sphinx.ext.napoleon", "m2r2"]
+
+autodoc_default_options = {
+    "members": True,
+    "show-inheritance": True,
+    "special-members": "__init__",
+    "undoc-members": True,
+}
+
+
+# Add any paths that contain templates here, relative to this directory.
+templates_path = ["_templates"]
+
+# The suffix(es) of source filenames.
+# You can specify multiple suffix as a list of string:
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = ".rst"
+
+# The master toctree document.
+master_doc = "index"
+
+# The name of the Pygments (syntax highlighting) style to use.
+pygments_style = "sphinx"
+
+# List of patterns, relative to source directory, that match files and
+# directories to ignore when looking for source files.
+# This pattern also affects html_static_path and html_extra_path.
+exclude_patterns = []
+
+
+# -- Options for HTML output ----------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = "sphinx_rtd_theme"
+
+# Theme options are theme-specific and customize the look and feel of a theme
+# further.  For a list of options available for each theme, see the
+# documentation.
+#
+# html_theme_options = {}
+
+
+# Add any paths that contain custom static files (such as style sheets) here,
+# relative to this directory. They are copied after the builtin static files,
+# so a file named "default.css" will overwrite the builtin "default.css".
+html_static_path = ["_static"]
+
+html_context = {
+    "css_files": [
+        "_static/theme_overrides.css",  # override wide tables in RTD theme
+        "_static/schema-page.css",  # Add css from jschemer
+    ]
+}
+
+html_sidebars = {"**": ["globaltoc.html", "relations.html", "sourcelink.html", "searchbox.html"]}
--- a/docs/source/netutils/index.rst
+++ b/docs/source/netutils/index.rst
@@ -1,0 +1,17 @@
+******************
+Netutils Functions
+******************
+
+.. toctree::
+   :maxdepth: 2
+
+   asn/index
+   configs/index
+   dns/index
+   interface/index
+   ip/index
+   mac/index
+   password/index
+   ping/index
+   route/index
+   vlan/index
--- a/netutils/mac.py
+++ b/netutils/mac.py
@@ -1,12 +1,14 @@
 """Functions for working with MAC addresses."""
 
 import re
+from functools import wraps
 from .constants import MAC_CREATE, MAC_REGEX
 
 
 def _valid_mac(func):
     """Decorator to validate a MAC address is valid."""
 
+    @wraps(func)
     def decorated(*args, **kwargs):
         if kwargs.get("mac"):
             mac = kwargs.get("mac")
--- a/tasks.py
+++ b/tasks.py
@@ -268,3 +268,30 @@ def check_pypi_version(context, name=PROJECT_NAME, version=PROJECT_VERSION):
     print(f"The version {version} does not exist on PyPI.")
     print("The version can be released.")
     sys.exit(0)
+
+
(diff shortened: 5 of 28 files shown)


docs/requirements.txt
```diff
@@ -0,0 +1 @@
+m2r2==0.2.7
```
itdependsnetworks, line 1: Should be added to poetry development requirements.

docs/source/conf.py
```diff
@@ -0,0 +1,101 @@
...
+# pylint: disable=W,C
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath("../.."))
+
+
+# -- Project information -----------------------------------------------------
+
+project = "netutils"
```
itdependsnetworks, line 21: Can we import from the project.toml file this information? That will keep it DRY.

docs/source/conf.py
```diff
@@ -0,0 +1,101 @@
...
+# You can specify multiple suffix as a list of string:
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = ".rst"
+
+# The master toctree document.
+master_doc = "index"
+
+# General information about the project.
+project = "netutils"
```
itdependsnetworks, line 63: seems redundant

docs/source/netutils/index.rst
```diff
@@ -0,0 +1,16 @@
+******************
+Netutils Functions
+******************
+
+.. toctree::
+   :maxdepth: 2
+
+   asn/index
```
itdependsnetworks, line 8: Where is config?

docs/tasks.py
```diff
@@ -0,0 +1,27 @@
...
+@task
+def html(context, sourcedir="source", builddir="build"):
+    """Creates html docs using sphinx-build command.
+
+    Args:
+        context (obj): Used to run specific commands
+        sourcedir (str, optional): Source directory for sphinx to use. Defaults to "source".
+        builddir (str, optional): Output directory for sphinx to use. Defaults to "build".
+    """
+    command = "sphinx-build" + sourcedir + builddir
```
itdependsnetworks, line 14: I guess I am missing something, but to my untrained eye it seems like this is missing whitespace.

docs/tasks.py
```diff
@@ -0,0 +1,27 @@
...
+
+@task
+def html(context, sourcedir="source", builddir="build"):
+    """Creates html docs using sphinx-build command.
+
+    Args:
+        context (obj): Used to run specific commands
+        sourcedir (str, optional): Source directory for sphinx to use. Defaults to "source".
+        builddir (str, optional): Output directory for sphinx to use. Defaults to "build".
+    """
```
itdependsnetworks, line 13: Good to add a print statement as well

netutils/mac.py
```diff
@@ -86,7 +88,6 @@ def mac_to_int(mac):
     return int(mac_normalize(mac), 16)
 
 
-@_valid_mac
```
itdependsnetworks: Why remove this?

itdependsnetworks:
Good stuff!

We should add a test or two. 
* Ensure it runs and compiles. 
* I think the end-line/start-line is great, but suspect that will move over time, can compare it against current standard.
* See if there is a way to verify that all files are considered.

docs/requirements.txt
qduk: I added this to poetry. I believe this also has to be in the requirements file for readthedocs. That's my understanding at least. I can verify that.

docs/source/conf.py
qduk: Updated.

docs/source/conf.py
qduk: Removed.

docs/source/netutils/index.rst
qduk: The config for `asn`? It lives in its own folder under `source/netutils/asn`.

docs/tasks.py
qduk: You are correct. Updated.

docs/tasks.py
qduk: Added.

netutils/mac.py
qduk: I removed it to test the autodoc feature of Sphinx. I missed putting it back. Updated.

qduk:
@itdependsnetworks I added a few tests. I'm not sure if they should live in the unit tests folder or not but they check the following,

- Sphinx builds without any errors 
- Checks that there is a folder in `docs/source/netutils` for a given python file
- Each folder contains an index.  

There is a [PR](https://github.com/miyakogi/m2r/pull/42) that would've allowed me to use the `start-before` and `end-after` directives which would've been easier to break the README into sections but it was never merged into the project. So the line directives seemed like the best next bet.

docs/source/netutils/index.rst
itdependsnetworks: like should there be an `config/index`?

docs/source/conf.py
```diff
@@ -12,13 +12,15 @@
...
+import toml
 
 sys.path.insert(0, os.path.abspath("../.."))
+toml_dict = toml.load("../../pyproject.toml")
 
 
 # -- Project information -----------------------------------------------------
 
-project = "netutils"
+project = toml_dict["tool"]["poetry"]["name"]
```
itdependsnetworks, line 23: For copyright and author as well.

docs/tasks.py
```diff
@@ -11,7 +11,8 @@ def html(context, sourcedir="source", builddir="build"):
         sourcedir (str, optional): Source directory for sphinx to use. Defaults to "source".
         builddir (str, optional): Output directory for sphinx to use. Defaults to "build".
     """
-    command = "sphinx-build" + sourcedir + builddir
+    print("Building html documentation...")
+    command = "sphinx-build " + sourcedir + " " + builddir
```
itdependsnetworks, line 15: We seem to use f-strings in our tasks.py

docs/source/conf.py
qduk: Updated

docs/tasks.py
qduk: Updated

docs/source/netutils/index.rst
qduk: Totally missed those. I added a general `configs` tab. We could break it up how the code is with each file having its own tab under config. Let me know what you think and I can change it if need be.

tasks.py
```diff
@@ -268,3 +268,29 @@ def check_pypi_version(context, name=PROJECT_NAME, version=PROJECT_VERSION):
...
+@task
+def html(context, sourcedir="docs/source", builddir="docs/build"):
+    """Creates html docs using sphinx-build command.
+
+    Args:
+        context (obj): Used to run specific commands
+        sourcedir (str, optional): Source directory for sphinx to use. Defaults to "source".
+        builddir (str, optional): Output directory for sphinx to use. Defaults to "build".
+    """
+    print("Building html documentation...")
```
itdependsnetworks, line 282: I think it makes sense to call clean_docs here, just ensure it is done idempotently.

tests/unit/test_docs.py
```diff
@@ -0,0 +1,43 @@
...
+
+UNDOCUMENTED_FILES = ["__init__", "constants", "lib_mapper", "protocol_mapper", "variables"]
+
+SPHINX_DIRECTORIES = [
+    {
+        "source_dir": "docs/source/",
+        "build_dir": "docs/build/",
+    }
+]
+
```
itdependsnetworks, line 15: Can you add a start/end line test here? Something to ensure that the current start and end line remain the same throughout updates. This will ensure that if the readme is modified, that the tests will fail. If the start or end line change, they simply need to modify the tests.

qduk: Added tests to check for that. Let me know what you think.

tasks.py
qduk: Updated

docs/source/conf.py
```diff
@@ -0,0 +1,99 @@
...
+extensions = ["sphinx.ext.autodoc", "sphinx.ext.napoleon", "m2r2"]
+
+autodoc_default_options = {
+    # "exclude-members": "__weakref__",
+    # "ignore-module-all": ,
+    # "imported-members": ,
+    # "inherited-members": ,
+    "members": True,
+    # "member-order": "bysource",
+    # "private-members": ,
```
jeffkala, line 45: Does it make sense to keep these?, being that they're just commented out?

jeffkala (approved):

docs/source/conf.py
qduk: Took them out. Good call. I believe Sphinx makes it that way when you start with it's template. Thanks Jeff.

tests/unit/test_docs.py
```diff
@@ -0,0 +1,87 @@
...
+@pytest.mark.parametrize("data", MODULE_FOLDERS)
+def test_folders_contain_index(data):
+    doc_module_folders = glob.glob(data["doc_module_folder"])
+
+    for folder in doc_module_folders:
+        assert "index.rst" in os.listdir(folder)
+
+
+def test_overview_section_start_end_lines(load_readme_into_lines):  # pylint: disable=redefined-outer-name
```
itdependsnetworks, line 52: Can we just create a data structure, and a single test. Something like 

```
[
  {
    "overview": {
      "start-line": 10,
      "start-val": "# netutils\n",
      "end-line": 32,
      "end-value": "* VLANs - Provide the ability to convert configuration into lists or lists into configuration.\n"
    }
  },
  {
    "attribution": {
      "start-line": "NN",
      "start-val": "TEXT",
      "end-line": "NN",
      "end-value": "TEXT"
    }
  }
]
```

Even better would be if you can dynamically find the line numbers, that are shown here, as an example:
```
   :start-line: 131   :end-line: 178
```

qduk: Updated. I created an issue for parsing the `start` and `end` line so I can get that handled down the road.

PR #9, opened 2021-05-14
Create and order interface lists

qduk:


netutils/interface.py
```diff
@@ -1,8 +1,89 @@
...
+
+    Returns:
+        int: Sum of number in interface tail with each one getting a priority of magnitude 10 from right to left.
+
+    Example:
+        >>> from netutils.interface import interface_tail_addition
+        >>> interface_tail_addition("1/0/6")
+        106
+        >>> interface_tail_addition("2/3/6")
+        236
```
glennmatthews, line 21: Does this fail in unexpected ways if any of the numbers are non-single-digits? (`0/0/48`, `10/0/16`)

dgjustice: It doesn't appear to:
```python
>>> interface_tail_addition("42/233/6123412")
6129942
>>>
```

glennmatthews: That would seem unexpected to me. :-) Put another way, shouldn't [`0/0/1`, `0/0/48`, `0/1/1`] sort into that given order? But these would evaluate as `1`, `48`, and `11`, resulting in a sorted order of [`0/0/1`, `0/1/1`, `0/0/48`]

dgjustice:
```python
>>> from netutils.interface import list_alphabetical
>>> ["Fa0/0/1", "Fa0/0/48", "Fa0/1/1"]
['Fa0/0/1', 'Fa0/0/48', 'Fa0/1/1']
>>>```

File: netutils/interface.py [hunk 2]
```diff
@@ -66,6 +147,68 @@ def canonical_interface_name(interface, addl_name_map=None, verify=False):
...
+        raise ValueError("Order must be set to use reverse.")
+
+    if order and order not in INTERFACE_LIST_ORDERING_OPTIONS.keys():
+        raise ValueError(f"{order} is not one of the supported orderings")
+
+    if isinstance(addl_name_map, dict):
+        name_map.update(addl_name_map)
+
+    for interface in interfaces:
+        interface_type, interface_number = split_interface(interface)
```
itdependsnetworks, line 195: I was expecting to see a call to `canonical_interface_name` here, any reason to not leverage the existing function?

netutils/interface.py
```diff
@@ -1,8 +1,89 @@
...
+    contained_interfaces = set()
+
+    for interface in interfaces:
+        interface_name = split_interface(interface)[0]
+        contained_interfaces.add(interface_name)
+
+    for interface_name in contained_interfaces:
+        new_list = [x for x in interfaces if split_interface(x)[0] == interface_name]
+        if new_list:
+            separate_interfaces_list.append(new_list)
```
dgjustice, line 71: I sent you a starter snippet to simplify this a bit.  Leaving a note for others.  I'll poke at it some more in the morning.

netutils/interface.py
glennmatthews:
```python
>>> from netutils.interface import list_alphabetical
>>> list_alphabetical(["Fa0/0/1", "Fa0/0/48", "Fa0/1/1"])
['Fa0/0/1', 'Fa0/1/1', 'Fa0/0/48']
```

micahculpepper: If the goal is to be able to sort interface IDs, that can be accomplished with tuples:
```python
>>> interfaces = [(0,0,1), (0,0,48), (0,1,1)]
>>> sorted(interfaces)
[(0, 0, 1), (0, 0, 48), (0, 1, 1)]
```
What's the advantage of integers?

qduk: @micahculpepper That is awesome. I didn't even think of that. Thanks for sharing. I'm going to update my code to use that.

itdependsnetworks: That is a great one!! A few more use use cases to consider though. 
`["Fa0/0/2", "Fa0/0/1", "Fa0/0/1.100", "Fa0/0/2.200"]`
`["Gi0/1", "Gi0/0/1", "Gi0/0/2" ]`
`["Gi1/0/1", "Te1/0/1", "Gi1/0/2" ]`

I believe these are all real world use cases.

dgjustice:
```python
sorted(map(
    lambda iface: (
        lambda tup: (tup[0], *re.split(r"[./]", tup[1]))
    )(split_interface(iface)),
    ifaces
))
```
That works per Micah's recommendation, but you lose the context of the separator.  The reason this is a problem is that "Fa0/1" has lower precedence over "Fa0.1".  Domain logic rears its head on this one.

dgjustice: Bad habit, here it is in "real" python.  :)
```python
out = []
for iface in ifaces:
    head, tail = split_interface(iface)
    tup = (head, *re.split(r"[./]", tail))
    out.append(tup)

print(list(sorted(out)))
```

dgjustice:
@qduk, a friend needed a solution to this very problem today, so I put something together that I believe addresses the issues we found along the way.  Basically, it separates characters into classes with weights and does a 2-stage sort.  Feel free to copy/paste it to your heart's content.

https://gist.github.com/dgjustice/84117af73541728ece78cb33c6d37ecb

qduk:
@dgjustice Finally had a chance to dive into this. There was a lot I didn't understand at first but what you wrote is super cool! One issue I ran into is that it sorts interface without "/" or "." not in numerical order. For example, if I use the test data of 
`test_data = ['Port-channel40', 'Port-channel160', "Loopback99", "Loopback101", "Loopback98", "Loopback102"]` I get 
```
Loopback101
Loopback102
Loopback98
Loopback99
Port-channel160
Port-channel40
```

I ended up solving that by sorting the list in place using 
```
for k,v in groupby(test_data, lambda x: "/" not in x and "." not in x):
    if k == True:
        for r,s in groupby(v, lambda x: x.rstrip("0123456789")):
            s = list(s)
            first_index = test_data.index(s[0])
            last_index = test_data.index(s[-1])
            test_data[first_index:last_index+1] = sorted(test_data[first_index:last_index+1], reverse=True)
```
I'm sure there is a way to do it in the iteration of the tree but the above method seemed to work. 

I also ran into an issue where it seemed to be skipping interfaces. When I used this test data
`["Gi1/0/1", "Gi1/0/3", "Gi1/0/3.100", "Gi1/0/2", "Gi1/0/2.50", "Po40", "Po160", "Lo10"]`
 I get this back. 
 ```
Gi1/0/1
Gi1/0/2.50
Gi1/0/3.100
Lo10
Po160
Po40
```
I'm not really sure why it's missing the `Gi1/0/2` and `Gi1/0/3`. That's what I'm trying to look into now. Any help/advice would be appreciated! Thanks @dgjustice

**Comment by dgjustice** [comment]
@qduk I'll submit my own PR with tests.

File: netutils/interface.py (hunk 2 above)
Comment (qduk) [reply netutils/interface.py]: Good point. Cleaned it up a bit.

File: netutils/interface.py [hunk 4]
```diff
@@ -127,6 +128,65 @@ def canonical_interface_name(interface, addl_name_map=None, verify=False):
...
+    name_map = {}
+    name_map.update(BASE_INTERFACES)
+    canonical_interface_list = []
+    no_match_list = []
+
+    if reverse and not order:
+        raise ValueError("Order must be set to use reverse.")
+
+    if order and order not in INTERFACE_LIST_ORDERING_OPTIONS.keys():
+        raise ValueError(f"{order} is not one of the supported orderings")
```
jeffkala, line 170: Identical to L468-472, possibly make a small helper private function to check this?

tests/unit/test_interface.py
```diff
@@ -215,6 +474,34 @@ def test_canonical_interface_name_failure():
         interface.canonical_interface_name(**data)
```
jeffkala, line 476: Might just add two test to make sure the ValueErrors (from my previous comment) are checked.

jeffkala (changes requested):
Couple small ones.

netutils/interface.py
qduk: Great idea. I added a little function.

tests/unit/test_interface.py
qduk: I added a test to check that the `abbreviated_interface_name_list` function gives me the expected error. I had one for `canonical_interface_name_list` on line 488. Thanks Jeff!

jeffkala (approved):

netutils/interface.py
```diff
@@ -352,3 +428,94 @@ def sort_interface_list(interfaces: t.List[str]) -> t.List[str]:
...
+    interfaces, addl_name_map=None, addl_reverse_map=None, verify=False, order=None, reverse=None
+):
+    """Function to return a list of interface's abbreviated name.
+
+    Args:
+        interfaces (list): List of interface names you are attempting to abbreviate.
+        addl_name_map (dict, optional): A dict containing key/value pairs that updates the base mapping. Used if an OS has specific differences. e.g. {"Po": "PortChannel"} vs {"Po": "Port-Channel"}. Defaults to None.
+        addl_reverse_map (dict, optional): A dict containing key/value pairs that updates the abbreviated mapping. Defaults to None.
+        verify (bool, optional): Whether or not to verify the interface matches a known interface standard. Defaults to False.
+        order (str, optional): Determines what order the list of interfaces should be returned in. Defaults to None.
```
itdependsnetworks, line 446: It seems like the default would fail.

qduk: As in `order` being set to `None`? L470-471 only check that the order specified is valid if its defined and L502-503 only attempt to order the list if it's defined.

Issue #11, opened 2021-05-24
WLC Parser


jdrew82:
### Environment
* netutils version:  1.0.0

### Proposed Functionality
Add a parser for Cisco WLC devices.

### Use Case
As a network tech it'd be very helpful to be able to parse WLC devices for compliance purposes.

jdrew82:
@itdependsnetworks I've looked into this a bit and I think this is an error in the parser_map. It looks like it's currently using "cisco_aireos" mapping to the AIREOSParser class but if we're using the netmiko platforms then it should actually be cisco_wlc. I think the same could go for the bigip_f5 mapping.

itdependsnetworks:
Just seeing this now. We may have done this based on PYNTC. I am not sure what the pros/cons are, but let me rope in @jmcgill298

dgarros:
Is there a way to redefine this mapping per project if needed ? 
We won't be able to always have the name of the platform  matches the netmiko syntax.

jmcgill298:
It was based off Ansible.

@dgarros it is today, the `network_os` arg must be passed to the filter. In this example as it is today:

```
my_parser_map:
  cisco_wlc: "cisco_aireos"
```

`data | filter(network_os={{ my_parser_map[my_name_for_os] }}, *args)`

itdependsnetworks:
As Jacob mentioned, one can map the parser, this was also reflected in https://github.com/nautobot/nautobot-plugin-nornir/pull/20

PR #15, opened 2021-06-02
Fix WLC Parser

jdrew82:
This addresses issue #11.


netutils/config/compliance.py
```diff
@@ -6,7 +6,7 @@
     "arista_eos": parser.EOSConfigParser,
     "cisco_ios": parser.IOSConfigParser,
     "cisco_nxos": parser.NXOSConfigParser,
-    "cisco_aireos": parser.AIREOSConfigParser,
+    "cisco_wlc": parser.AIREOSConfigParser,
```
itdependsnetworks, line 9: Why is this preferred?

Issue #10, opened 2021-05-24
ASA Parser


jdrew82:
### Environment
* netutils version:  1.0.0

### Proposed Functionality
Add the ability to parse ASA configurations.

### Use Case
As a network/security technician it'd be great if ASA configurations were able to be parsed for compliance purposes.

itdependsnetworks:
Addressed in #16

PR #16, opened 2021-06-02
Add ASA Parser

jdrew82:
This is an initial attempt at a parser for ASAs to address #10.


netutils/config/parser.py
```diff
@@ -603,3 +603,78 @@ class JunosConfigParser(BaseSpaceConfigParser):
 
     comment_chars = []
     banner_start = []
+
+
+class ASAConfigParser(CiscoConfigParser, BaseSpaceConfigParser):
```
itdependsnetworks, line 608: CiscoConfigParser already inherits from BaseSpaceConfigParser, e.g. `class CiscoConfigParser(BaseSpaceConfigParser):`

This should suffice `class ASAConfigParser(CiscoConfigParser):`

itdependsnetworks:
Can you add tests/mocked examples for all of the compliance functions? e.g. in folder `https://github.com/networktocode/netutils/tree/develop/tests/unit/mock/config/compliance`

netutils/config/parser.py
jdrew82: This is due to using IOSConfigParser as an example and it inherits from those two classes. I will correct it.

jdrew82:
I'll work on getting those tests added today.

jeffkala (approved):
Looks good to me.

netutils/config/parser.py
```diff
@@ -603,3 +603,78 @@ class JunosConfigParser(BaseSpaceConfigParser):
...
+            ... [
+            ...     ConfigLine(config_line="interface Management0/0", parents=()),
+            ...     ConfigLine(config_line=" management-only", parents=("interface Management0/0",)),
+            ...     ConfigLine(config_line=" nameif Management", parents=("interface Management0/0",)),
+            ...     ConfigLine(config_line=" security-level 100", parents=("interface Management0/0",)),
+            ...     ConfigLine(config_line=" ip address 10.1.1.10 255.255.255.0", parents=("interface Management0/0",)),
+            ... ]
+            True
+        """
+        for line in self.generator_config:
```
itdependsnetworks, line 663: @jmcgill298 Can you give this a quick review? I get lost a little bit what this is doing, but it seems like there is a logical reason to overwrite this method. 

Please feel free to review entire PR as everything looks good to me and I am inclined to merge.

jdrew82: It was overwritten because the ASA does the banner differently. Instead of needing to parse it like it's a single line with a delimiter, you can just parse each line individually like normal since each line always starts with the banner keyword, ie:

```
banner motd <BANNER TEXT>
banner exec <BANNER TEXT>
```

itdependsnetworks: Thanks!! That makes sense.

jmcgill298 (approved):

itdependsnetworks (approved):

PR #25, opened 2021-06-12
move sphinx to dev dep, update docs, update metadata

itdependsnetworks:


tests/unit/test_docs.py
```diff
@@ -76,9 +85,7 @@ def test_folders_contain_index(data):
         assert "index.rst" in os.listdir(folder)
 
 
-@pytest.mark.parametrize("start_end", START_END_LINES, ids=[list(x.keys())[0] for x in START_END_LINES])
-def test_docs_start_end_lines(start_end, load_readme_into_lines):  # pylint: disable=redefined-outer-name
-    readme_data = load_readme_into_lines
-    section = list(start_end.keys())[0]
-    assert readme_data[start_end[section]["start_line"]] == start_end[section]["start_value"]
-    assert readme_data[start_end[section]["end_line"]] == start_end[section]["end_value"]
+@pytest.mark.parametrize("start_end", START_END_LINES)
```
itdependsnetworks, line 88: @qduk please verify this, I couldn't really understand then previous logic, but this seems to work as I would have wanted it to.

qduk: @itdependsnetworks Looks great to me. Adding the empty line at the beginning was smart. 
Can we add id's to the test like so?
`@pytest.mark.parametrize("start_end", START_END_LINES, ids=[section["name"] for section in START_END_LINES])`

qduk (approved):
Looks good!

jeffkala (approved):

PR #28, opened 2021-06-19
Add IPv6 functionality to ip module.

dgjustice:
Adds dual-stack functionality anywhere it was missing.  Please see commit comments for more details.  The one oddball is `cidr_to_netmask` because of the ambiguity in translating valid masks <= 32.


netutils/ip.py
```diff
@@ -167,22 +171,43 @@ def cidr_to_netmask(cidr):
     raise ValueError("Parameter must be an integer between 0 and 32.")
 
 
+def cidr_to_netmask6(cidr):
```
itdependsnetworks, line 174: nit naming, does `cidr_to_netmaskv6` make more sense?

netutils/ip.py
```diff
@@ -247,3 +272,25 @@ def get_usable_range(ip_network):
         lower_bound = str(net[1])
         upper_bound = str(net[-2])
     return f"{lower_bound} - {upper_bound}"
+
+
+def count_bits(i: int) -> int:
```
itdependsnetworks, line 277: Should this be a private method? 

Also, should avoid using single letter variable names.

netutils/ip.py
```diff
@@ -167,22 +171,43 @@ def cidr_to_netmask(cidr):
...
+
 def get_all_host(ip_network):
     """Given a network, return the list of usable IP addresses.
 
     Args:
         ip_network (str): An IP network in string format that is able to be converted by `ipaddress` library.
 
     Returns:
-        list: List of usable IP Addresses within network.
+        generator: Generator of usable IP Addresses within network.
```
itdependsnetworks, line 202: I can go either way on this, performance is obviously better as a generator, but just want to make sure it makes sense for the masses.

itdependsnetworks:
Thanks!! Mostly just a bunch of nit's, but appreciate the addition.

netutils/ip.py
dgjustice: I can yank this out if you like.

netutils/ip.py
dgjustice: Up to you.

jeffkala (approved):

itdependsnetworks:
Last nit for me, can you remove any single letter variables?

PR #31, opened 2021-06-29
interface_range_expansion function added

qduk:


netutils/interface.py
```diff
@@ -1,6 +1,73 @@
...
+        interface_pattern (str): The string pattern that will be parsed to create the list of interfaces.
+
+    Returns:
+        list: Contains the expanded range of values.
+
+    Example:
+        >>> from netutils.interface import interface_range_expansion
+        >>> interface_range_expansion("Gi0/[1-4]")
+        ['Gi0/1', 'Gi0/2', 'Gi0/3', 'Gi0/4']
+        >>> interface_range_expansion("[ge,xe]0/[0-2]")
```
itdependsnetworks, line 55: I don't know that we need to support this pattern, `[ge,xe]` it terms of allowing characters to be expanded. It seems like it may be more trouble than it is worth.

itdependsnetworks:
What about 
```
import re
import itertools
def interface_range_expansion(interface_pattern):    

    # from https://rosettacode.org/wiki/Range_expansion#Python
    def _rangeexpand(txt):
        lst = []
        for r in txt.split(','):
            if '-' in r[1:]:
                r0, r1 = r[1:].split('-', 1)
                lst += range(int(r[0] + r0), int(r1) + 1)
            else:
                lst.append(int(r))
        return lst

    # from https://stackoverflow.com/questions/4628290/pairs-from-single-list
    def _pairwise(t):
        it = iter(t)
        return list(zip(it,it))

    match_pattern = r"(\[(?:\d|,|-)+\])"
    re_compiled = re.compile(match_pattern)
    # Use case when sent without an actual range, e.g. Gi1
    if not re_compiled.search(interface_pattern):
        return [interface_pattern]

    cartesian_list = []
    interface_constant = [0]
    for match in re_compiled.finditer(interface_pattern):
        interface_constant.append(match.start())
        interface_constant.append(match.end())
        cartesian_list.append(_rangeexpand(match.group()[1:-1]))

    # Use case where interface range is at the end, e.g. Gi[1-10]
    if len(interface_pattern) == interface_constant[-1]:
        interface_constant.pop()
    else:
        interface_constant.append(len(interface_pattern))

    interface_constant_out = _pairwise(interface_constant)

    expanded_interfaces = []
    for element in itertools.product(*cartesian_list):
        current_interface = ""
        for count, item in enumerate(interface_constant_out):
            current_interface += interface_pattern[item[0]:item[1]]
            current_interface += str(element[count])
        expanded_interfaces.append(current_interface)
    return expanded_interfaces
```

qduk:
@itdependsnetworks I like it. I also learned about some new things going through the code. I did change some variable names to conform to pylint and hopefully improve readability.

PR #33, opened 2021-07-17
add interface_range_compress function

viktorkertesz:
Dear NTC,

I thought it was useful to have this function in this library.

Regards,
Viktor

### DIFF
--- a/netutils/interface.py
+++ b/netutils/interface.py
@@ -1,6 +1,8 @@
 """Functions for working with interface."""
 import re
 import itertools
+from typing import Union, List, Tuple
+from operator import itemgetter
 from .constants import BASE_INTERFACES, REVERSE_MAPPING
 
 
@@ -172,3 +174,128 @@ def abbreviated_interface_name(interface, addl_name_map=None, addl_reverse_map=N
         raise ValueError(f"Verify interface on and no match found for {interface}")
     # If abbreviated name lookup fails, return original name
     return interface
+
+
+def interface_range_generator(interfaces: Union[str, List[str]] = None,
+                              prefix: str = "interface range ",
+                              max_ranges: int = 5) -> List[str]:
+    """Function which takes interfaces and return interface ranges.
+    Ranges are based on last interface part (aka ports).
+    Whitespace characters are ignored in the input.
+
+    Example:
+        >>> interface_range_generator("Gi1/0/1 Gi1/0/2 Gi1/0/4")
+        'interface range Gi1/0/1-2,Gi1/0/4'
+        >>> interface_range_generator(["Gi1/0/1 Gi1/0/2","Gi1/0/3,Gi1/0/4"])
+        'interface range Gi1/0/1-4'
+
+    Args:
+        prefix: prefix string used in each line of the output
+        interfaces: interfaces in a line or in a list of lines
+        max_ranges: maximum number of ranges in one line in the output
+
+    Returns:
+        lines of interface range commands (or ranges with specified prefix)
+    """
+
+    def to_port(port_in: Tuple[str, int, int, int, int]):
+        """Assemble exploded port_in
+        Separator is `/`. -1 value means that value is not used.
+        Only physical ports are supported.
+        Example: Gi, 1, 0, 1, -1 will become: Gi1/0/1
+
+        Args:
+            port_in: tuple of port name and fex, chasses, module, interface numbers
+
+        Returns:
+            string of assembled interface
+        """
+        out = port_in[0]
+        if port_in[1] >= 0:
+            out += str(port_in[1])
+            if port_in[2] >= 0:
+                out += "/" + str(port_in[2])
+                if port_in[3] >= 0:
+                    out += "/" + str(port_in[3])
+                    if port_in[4] >= 0:
+                        out += "/" + str(port_in[4])
+        return out
+
+    ports = []  # collect exploded interfaces
+    port_regex = re.compile(r"(?i)([a-z]+)([0-9]+)(?:/([0-9]+))?(?:/([0-9]+))?(?:/([0-9]+))?")
+    output = []
+    if interfaces is None:
+        return output
+    ports_in = [interfaces] if type(interfaces) is str else interfaces
+    # read all lines and explode all interfaces as preparation for sorting
+    for line in ports_in:
+        matches = port_regex.findall(line)
+        for match in matches:
+            port_name = match[0]
+            port1 = int(match[1]) if len(match[1]) > 0 else -1
+            port2 = int(match[2]) if len(match[2]) > 0 else -1
+            port3 = int(match[3]) if len(match[3]) > 0 else -1
+            port4 = int(match[4]) if len(match[4]) > 0 else -1
+            ports.append((port_name, port1, port2, port3, port4))
+
+    # sort exploded interface data in order to prepare for finding ranges
+    ports = sorted(ports, key=itemgetter(0, 1, 2, 3, 4))  # Sort interfaces
+    if not ports:  # could not read interfaces from input
+        return []
+    range_start = ports[0]  # contains interface range start (Gi0/0)
+    range_index = 1  # counts number of ranges on line (gi0/0-1,g0/3-5 = 2)
+    current_port = range_start
+    last_port_index = 1
+    outline = "%s%s" % (prefix, to_port(range_start))
+    i = 0  # range counter
+    for port in ports[1:]:
+        # interface name
+        if range_start[0] != port[0]:
+            if i > 0:  # it's a range
+                outline += "-%d" % current_port[last_port_index]
+                i = 0  # reset range counter
+            if range_index >= 5:
+                range_index = 1
+                output.append(outline)
+                outline = "%s%s" % (prefix, to_port(port))
+            else:
+                range_index += 1
+                outline += ", %s" % to_port(port)
+            range_start = port
+            current_port = port
+            continue  # move ahead for next port
+
+        # interface numbers
+        for port_index in range(1, 5):
+            if current_port[port_index] == port[port_index]:
+                pass
+            elif port[port_index] == (current_port[port_index] + 1) and port_index == 4:
+                i += 1
+                current_port = port
+                last_port_index = port_index
+                break
+            elif port[port_index] == (current_port[port_index] + 1) and port[port_index + 1] < 0:  # port_index < 4
+                # check if we found the last sane index
+                i += 1
+                current_port = port
+                last_port_index = port_index
+                break
+            else:
+                if i > 0:  # it's a range
+                    outline += "-%d" % current_port[last_port_index]
+                    i = 0  # reset range counter
+                if range_index >= max_ranges:
+                    range_index = 1
+                    output.append(outline)
+                    outline = "%s%s" % (prefix, to_port(port))
+                else:
+                    range_index += 1
+                    outline += ", %s" % to_port(port)
+                range_start = port
+                current_port = port
+                break  # move ahead for next port
+
+    if i > 0:
+        outline += "-%d" % current_port[last_port_index]
+    output.append(outline)
+    return output
--- a/netutils/interface.py
+++ b/netutils/interface.py
@@ -2,6 +2,7 @@
 import re
 import itertools
 from typing import Union, List, Tuple
+from collections import namedtuple
 from operator import itemgetter
 from .constants import BASE_INTERFACES, REVERSE_MAPPING
 
@@ -176,7 +177,7 @@ def abbreviated_interface_name(interface, addl_name_map=None, addl_reverse_map=N
     return interface
 
 
-def interface_range_generator(
+def interface_range_compress(
     interfaces: Union[str, List[str]] = None, prefix: str = "interface range ", max_ranges: int = 5
 ) -> List[str]:
     """Function which takes interfaces and return interface ranges.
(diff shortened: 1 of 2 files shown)


netutils/interface.py
```diff
@@ -172,3 +175,148 @@ def abbreviated_interface_name(interface, addl_name_map=None, addl_reverse_map=N
...
+
+    ports = []  # collect exploded interfaces
+    # case insensitive port parsing. We do a hard assumption that we only have ports in the input.
+
+    output = []
+    if interfaces is None:
+        return []
+    # read all lines and explode all interfaces as preparation for sorting
+    for line in [interfaces] if isinstance(interfaces, str) else interfaces:
+        # matches = port_regex.findall(line)
```
qduk, line 246: Can this be removed? Might have been part of the development process.

qduk:
@viktorkertesz Thank you for the PR. Can you add some tests that check for changes in the `prefix` and `max_ranges`? Something along the lines of 
```
    {
        "sent": {
            "interfaces": "vlan1, vlan2, vlan3, vlan4, vlan5, vlan6, vlan8, vlan10, vlan20",
            "prefix": "test prefix",
            "max_range": 20
        }, 
        "received": {
            
        }
    }
    ```

File: netutils/interface.py [hunk 2]
```diff
@@ -172,3 +175,148 @@ def abbreviated_interface_name(interface, addl_name_map=None, addl_reverse_map=N
         raise ValueError(f"Verify interface on and no match found for {interface}")
     # If abbreviated name lookup fails, return original name
     return interface
+
+
+def interface_range_compress(
+    interfaces: Union[str, List[str]] = None, prefix: str = "interface range ", max_ranges: int = 5
```
qduk, line 181: Can we default `interfaces` to an empty string to be in line with our typing? or use Optional?

netutils/interface.py
```diff
@@ -172,3 +175,148 @@ def abbreviated_interface_name(interface, addl_name_map=None, addl_reverse_map=N
...
+        """
+        out = port_in.iface_name
+        if port_in.module1 >= 0:
+            out += str(port_in.module1)
+            if port_in.module2 >= 0:
+                out += "/" + str(port_in.module2)
+                if port_in.module3 >= 0:
+                    out += "/" + str(port_in.module3)
+                    if port_in.module4 >= 0:
+                        out += "/" + str(port_in.module4)
```
qduk, line 235: What are your thoughts on doing something like this? Same idea but uses less `ifs` and uses `f-strings`.
```
        for idx, module in enumerate(port_in[1:]):
            if module >= 0 and idx == 0:
                out += str(module)
            elif module >= 0:
                out += f"/{str(module)}"
                ```

File: netutils/interface.py (hunk 1 above)
Comment (viktorkertesz) [reply netutils/interface.py]: Cleaned up in [last commit](https://github.com/networktocode/netutils/pull/33/commits/8f9a51aaa0357c92a89eb88a5ae84c2797c2c359#diff-e880d02b86528af00499f4d00b3d768decbd9ff9d6c5921ee90479c468e5e043L246)

File: netutils/interface.py (hunk 3 above)
Comment (viktorkertesz) [reply netutils/interface.py]: refactored to your suggestion in [last commit](https://github.com/networktocode/netutils/pull/33/commits/8f9a51aaa0357c92a89eb88a5ae84c2797c2c359#diff-e880d02b86528af00499f4d00b3d768decbd9ff9d6c5921ee90479c468e5e043R228-R232)

File: netutils/interface.py (hunk 2 above)
Comment (viktorkertesz) [reply netutils/interface.py]: Changed to Optional in [last commit](https://github.com/networktocode/netutils/pull/33/commits/8f9a51aaa0357c92a89eb88a5ae84c2797c2c359#diff-e880d02b86528af00499f4d00b3d768decbd9ff9d6c5921ee90479c468e5e043R181)

**Comment by jeffkala** [comment]
@viktorkertesz thanks for the clean ups.  Looks like `pylint` is failing on the CI build now.  Please fix the pylint errors.  Thanks.

**Comment by viktorkertesz** [comment]
@jeffkala , can you please help me with this? I tried to add noqa exceptions for pylint, but it keeps complaining.

**Comment by jeffkala** [comment]
> @jeffkala , can you please help me with this? I tried to add noqa exceptions for pylint, but it keeps complaining.

Pylint uses `pylint: disable=<code>` syntax to ignore

**Comment by qduk** [comment]
Hey @viktorkertesz , 
 We discussed this PR internally. We think it's a great idea to include a more fundamental version of this in the library. A few changes we suggest:
 
 1. Remove the `prefix` from the function. While this may come in handy sometimes, overall we think it'd be better to deal with just the numbers and leave appending a prefix up to the developer. 
 2. Allowing for multiple interfaces in one list element (i.e interface_range_compress(["Gi1/0/1 Gi1/0/2","Gi1/0/3,Gi1/0/4"])) can lead to confusion. We think it'd be best to have each list item represent a single interface. 
 3. Changing the code to reflect the suggestion above should also eliminate the need for the `max_range` argument. A developer should be able to put 0,1, or n number of items into the function.
 
 If you have any questions let us know.

**Comment by viktorkertesz** [comment]
Hi Adam, team,

As I understood, you want a much simpler approach and you don’t want to stick to Cisco defaults. My original use case was to copy/paste ‘show vlan br’ output and feed it to this function. That gave me a ready-to-use output for reconfiguring those interfaces easily. I can definitely do what you require, but then all users (including Cisco users) have to implement an additional layer if they want the same functionality as before. Maybe I can add a git Gist which implements that layer.

“Allowing for multiple interfaces in one list element (i.e interface_range_compress(["Gi1/0/1 Gi1/0/2","Gi1/0/3,Gi1/0/4"])) can lead to confusion. We think it'd be best to have each list item represent a single interface. “
I will keep the original functionality to keep the regex for recognizing all interfaces. I’ll update docs for users to use only 1 interface per list item.

“Changing the code to reflect the suggestion above should also eliminate the need for the max_range argument. A developer should be able to put 0,1, or n number of items into the function. “
My opinion is that this is a loss of functionality but I’ll do that. Gist will implement the original behavior.

Let me work on this and I’ll get back to you! Thank you for your time and considering my code!

**Comment by realvitya** [comment]
Hi Team,

I incorporated what you asked. Hope this more base solution is better :)

Viktor

**Comment by jeffkala** [comment]
Hoping to take a look at this in the next few days.

**Comment by jeffkala** [comment]
@viktorkertesz does the range command allow for module spanning ranges? I don't believe it does but don't have anything to test on.
Mean `interface range gi1/0/1-gi1/1/0`
Pretty sure it doesn't allow for this so I'm thinking we can accomplish the range compress by simply looking at the last module number and doing the collapse.

**Comment by jeffkala** [comment]
Something like this does work, and utilizes some of the functionality already built into this library.
```
from netutils.interface import _split_interface_tuple, sort_interface_list
from netutils.vlan import vlanlist_to_config

def create_common(interface_list):
    """Temp"""
    result_dict = {}
    final_result_list = []
    sorted_ints = [_split_interface_tuple(x) for x in sort_interface_list(interface_list)]
    current_match = sorted_ints[0][0:-1]
    for interface in sorted_ints:
        if interface[0:-1] == current_match:
            a = "".join([x.val for x in current_match])
            if result_dict.get(a):
                result_dict[a] += [int(interface[-1].val)]
            else:
                result_dict[a] = [int(interface[-1].val)]
        else:
            current_match = interface[0:-1]
            result_dict["".join([x.val for x in current_match])] = [int(interface[-1].val)]
    for k, v in result_dict.items():
        # vlanlist_to_config collapses ranges but if its 3 in a row or longer.
        [final_result_list.append(f'{k}{r}') for r in vlanlist_to_config(v)[0].split(',')]
    return final_result_list


ints = ["Gi1/1/9", "Gi1/0/1", "Gi1/0/2", "Gi1/0/3", "Gi1/0/5", "Gi1/1/1", "Gi1/1/2", "Gi1/1/10", "Gi1/1/11"]
result = create_common(ints)
print(result)

```
One caveat is it only merges if 3 consecutive numbers are found. Mean 1,2 will not be Gi1/1/1-2 they will be each their own, but anything > 3 will be.

**Comment by viktorkertesz** [comment]
> One caveat is it only merges if 3 consecutive numbers are found. Mean 1,2 will not be Gi1/1/1-2 they will be each their own, but anything > 3 will be.

Just because I see you like list comprehension orgy, here is the solution for the <3 cases :)
```python
from netutils.interface import _split_interface_tuple, sort_interface_list


def create_common(interface_list):
    """Temp"""
    result_dict = {}
    final_result_list = []
    sorted_ints = [_split_interface_tuple(x) for x in sort_interface_list(interface_list)]
    current_match = sorted_ints[0][0:-1]
    for interface in sorted_ints:
        if interface[0:-1] == current_match:
            a = "".join([x.val for x in current_match])
            if result_dict.get(a):
                result_dict[a] += [int(interface[-1].val)]
            else:
                result_dict[a] = [int(interface[-1].val)]
        else:
            current_match = interface[0:-1]
            result_dict["".join([x.val for x in current_match])] = [int(interface[-1].val)]
    for k, v in result_dict.items():
        # ultimate interface number compressor:
        [final_result_list.append(f'{k}{r}') for r in [f"{s}-{e}" for s, e in
           zip([x for x in v if x - 1 not in v and x + 1 in v],
               [x for x in v if x - 1 in v and x + 1 not in v and x not in
                  [x for x in v if x - 1 not in v and x + 1 in v]])] +
                  [f"{x}" for x in v if x - 1 not in v and x + 1 not in v]
         ]
    return final_result_list

ints = ["Gi1/1/9", "Gi1/0/1", "Gi1/0/2", "Gi1/0/3", "Gi1/0/5", "Gi1/1/1", "Gi1/1/2", "Gi1/1/10", "Gi1/1/11"]
result = create_common(ints)
print(result)

result:
['Gi1/0/1-3', 'Gi1/0/5', 'Gi1/1/1-2', 'Gi1/1/9-11']
```
To be honest, this is ugly and inefficient. I prefer my original solution. I just had some fun finding and using this:
```python
data = [1, 3, 5, 6, 7, 9]

starts = [x for x in data if x-1 not in data and x+1 in data]
ends = [x for x in data if x-1 in data and x+1 not in data and x not in starts]
singles = [f"{x}" for x in data if x-1 not in data and x+1 not in data]
print(list(zip(starts, ends)) + singles)
```

jeffkala:
I'm not married to my solution, just want to reuse functions that already exist.  I think my purpose for what I posted was more of a general second option/starting point.  I'll let @itdependsnetworks and @qduk weigh in as well.  I think the original implementation needs to be cleaned up and simplified to make sure maintainability can be achieved.

viktorkertesz:
@jeffkala , okie, I didn't mean your solution is ugly, I referred to my list comprehension block.
Actually your approach is nice but I would use a simpler code to do the ranges instead of `vlanlist_to_config`. Some nice solutions are [here](https://stackoverflow.com/questions/2154249/identify-groups-of-continuous-numbers-in-a-list).

viktorkertesz:
How about this?
```python
from netutils.interface import _split_interface_tuple, sort_interface_list
from operator import itemgetter
from itertools import groupby

def create_common(interface_list):
    """Temp"""
    result_dict = {}
    final_result_list = []
    sorted_ints = [_split_interface_tuple(x) for x in sort_interface_list(interface_list)]
    current_match = sorted_ints[0][0:-1]
    for interface in sorted_ints:
        if interface[0:-1] == current_match:
            a = "".join([x.val for x in current_match])
            if result_dict.get(a):
                result_dict[a] += [int(interface[-1].val)]
            else:
                result_dict[a] = [int(interface[-1].val)]
        else:
            current_match = interface[0:-1]
            result_dict["".join([x.val for x in current_match])] = [int(interface[-1].val)]
    for k, v in result_dict.items():
        ranges = [list(map(itemgetter(1), g)) for k, g in groupby(enumerate(v), lambda x:x[0]-x[1])]
        [final_result_list.append(f'{k}{r[0]}'+(f'-{r[-1]}' if len(r)>1 else '')) for r in ranges]
    return final_result_list

ints = ["Gi1/1/9", "Gi1/0/1", "Gi1/0/2", "Gi1/0/3", "Gi1/0/5", "Gi1/1/1", "Gi1/1/2", "Gi1/1/10", "Gi1/1/11"]
result = create_common(ints)
print(result)

result:
['Gi1/0/1-3', 'Gi1/0/5', 'Gi1/1/1-2', 'Gi1/1/9-11']
```

jeffkala:
I think this is looking much better and easier to maintain, if you can just update your PR with it and if possible update docstring, change the single letter variables and make them descriptive, and run the test.

viktorkertesz:
I used the above code to rewrite the function. Also renamed variables where it was sane. I added the `_ranges_in_list` helper function you mentioned and updated tests.

netutils/interface.py
```diff
@@ -519,3 +518,62 @@ def _check_order_option_exists(order):
...
+    """Function which takes interfaces and return interface ranges.
+
+    Whitespace and special characters are ignored in the input. Input must contain only interfaces,
+    there is no check against correct interface names! Also interface names must use the same abbreviation!
+    E.g. Gi =! GigabitEthernet
+
+    Example:
+        >>> interface_range_compress(["Gi1/0/1", "Gi1/0/2", "Gi1/0/3", "Gi1/0/5"])
+        ['Gi1/0/1-3', 'Gi1/0/5']
+
```
qduk, line 549:
```suggestion
    Example:
        >>> interface_range_compress(["Gi1/0/1", "Gi1/0/2", "Gi1/0/3", "Gi1/0/5"])
        ['Gi1/0/1-3', 'Gi1/0/5']
        >>> interface_range_compress(["Gi0/1", "Gi0/2", "Gi0/4", "Gi1/0", "Gi1/1"])
        ["Gi0/1-2", "Gi0/4", "Gi1/0-1"]
```

PR #34, opened 2021-07-28
Add interface sorting support.

dgjustice:
@qduk I have refactored the gist I sent (thank you for pointing out how broken that was! :grimacing:).  I finally put together a proper PR with all the test cases I could find.  This is a bit more complexity, but taking the "character class" approach allows future modifications to completely customize the sorting logic based on arbitrary groups of characters.  Sorts happen first by `CharacterClass.weight`, and secondly by whatever logic you like, for example, casting to `int` for the `CCInt` class.

### DIFF
--- a/netutils/interface.py
+++ b/netutils/interface.py
@@ -1,6 +1,10 @@
 """Functions for working with interface."""
-import re
 import itertools
+import re
+import typing as t
+from abc import ABC, abstractmethod, abstractproperty
+from functools import total_ordering
+
 from .constants import BASE_INTERFACES, REVERSE_MAPPING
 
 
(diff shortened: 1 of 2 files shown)


qduk (changes requested):
@dgjustice gotta say, this is pretty awesome stuff. I've just got one other note. 

Docs is picking up on your character classes and helper functions. I personally don't think those need to be in there. I am able to get them undocumented by making them private functions. I'll defer to @itdependsnetworks and @jeffkala to see if they agree or not. 

Thanks again @dgjustice. Once this is merged it'll allow me to build out my PR.

dgjustice:
Travis is being flaky again.  I think the last commit is clean (passes tests); the build failed on installing dependencies.

qduk (approved):
@jeffkala You make a good point. Never hurts to have the extra documentation. 
@dgjustice Thanks for adding the docstring example.

netutils/interface.py
```diff
@@ -172,3 +176,179 @@ def abbreviated_interface_name(interface, addl_name_map=None, addl_reverse_map=N
...
+    def weight(self) -> int:  # noqa: D102
+        return 30
+
+
+def _CCfail(*args):  # pylint: disable=C0103
+    """Helper to raise an exception on a bad character match."""
+    raise ValueError(f"unknown character '{args[0][0]}'.")
+
+
+def split_interface_tuple(interface: str) -> t.Tuple[CharacterClass, ...]:
```
itdependsnetworks, line 269: Can we mark split_interface_tuple and insert_nodes as private? I rather be explicit that these are not supported functions and subject to change.

dgjustice: Yes, no problem.

PR #40, opened 2021-08-18
Port to num

matt852:
Creates a port to number mapping for common ports.
Includes reverse lookup.
Also includes common port numbers to protocol mappings (tcp, udp, or tcp-udp).


jeffkala:
Hey @matt852 Might collide withw #35

Issue #29, opened 2021-06-25
Convenience method for Jinja2 filters


davidban77:
### Environment
* netutils version:  0.3.0 

### Proposed Functionality

This can be a great library for **jinja2** filters usage, since it provides no extra dependencies, hence is lightweight, functions are tested and it will be easy to have it imported and approved for your jinja2 environment.

Currently if you want to use it for jinja2 filters, you will need to import each function individually and add it to your filters environment, which can be cumbersome. So it would be great to have a convenience method to import the library as whole (or maybe as per module) and assign as a filter.

### Use Case

It can be used in Ansible and Nornir scenarios for example, giving the developer just some placeholders commands to set it up and having all the functions as available filters in the templates.

Lets say the python jinja2 setup is like:

```python
from jinja import Environment
from netutils import netutils_as_jinja2_filter

# Setup jinja2 environment
env = Environment(...)

# Assign netutils filter
env.filters["netutils"] = netutils_as_jinja2_filter
```

And on your templates you can use it like:

```
IP Address + 200 = {{ "192.168.0.1" | netutils("ip.ip_addition", 200) }}
```

To render:

```
IP Address + 200 = 192.168.0.201
```

PR #45, opened 2021-09-07
WIP jinja2 convenience function

qduk:
This is just a WIP. I want to get your guys opinion on this approach for a jinja2 convenience function. I would still need to build tests, give examples, update documentation, add error handling and add some sort of validation for parameter inputs. The usage is how David Flores mentioned in #29. You would import the convenience function and then register it as a filter.  Thoughts?


itdependsnetworks:
I was thinking something like
```
_JINJA_MAP = {
	"fqdn_to_ip": "netutils.dns.fqdn_to_ip",
	"is_fqdn_resolvable": "netutils.dns.is_fqdn_resolvable",
	etc...
}

def jinja2_convenience_function():
    result = {}
    for key, value in _JINJA_MAP.items():
        result[key] = import_module(value)

    return result
```


And example usage:
```
import jinja2

env = jinja2.Environment()
env.filters.extend(jinja2_convenience_function())

```

Then we can build a test that every function get's added to the _JINJA_MAP

qduk:
That's a good idea. Didn't know if using one filter or splitting it up would be best. Hence why I wanted to run it by you guys. I'll update this PR with the refactor, proper tests, etc.

qduk:
@itdependsnetworks thoughts? This presents each function as its own filter and dynamically finds the files needed. If a new folder was created the code would need updating but anything in the `root` or `config` directories will be included.

itdependsnetworks:
Would be good to get someone else to chime in (@jeffkala ?? ) I still personally prefer the explicit mapping. I also think this should help for documentation as well as naming. As example, `get_usable_range` by itself is not as descriptive, outside of the context of being an ip check. Perhaps we would change the name to `get_usable_range_ip` or similar.

jeffkala:
I like explicit mapping in this case.  I think its easier to read and understand the code that way.  Also would agree with Ken that the documentation would be easier.

qduk:
@jeffkala @itdependsnetworks I revamped this. Let me know what you guys think.

netutils/utils.py
```diff
@@ -0,0 +1,81 @@
+"""Utilities for the netutils library."""
+from importlib import import_module
+
+_JINJA2_FUNCTION_MAPPINGS = {
+    "asn_to_int": "asn.asn_to_int",
+    "config": {
```
itdependsnetworks, line 6: I don't understand this nesting

docs/source/netutils/utilities/index.rst
```diff
@@ -0,0 +1,65 @@
...
+    from jinja2.loaders import FileSystemLoader, PackageLoader
+    from jinja2 import Environment, PackageLoader, select_autoescape
+    from netutils.utils import jinja2_convenience_function
+
+    env = Environment(
+        loader=FileSystemLoader("templates"),
+        autoescape=select_autoescape()
+    )
+
+    for function_name, function_object in jinja2_convenience_function().items():
```
itdependsnetworks, line 37: I said extend, but I meant to say update, does this work? env.filters.update(jinja2_convenience_function())

netutils/utils.py
```diff
@@ -0,0 +1,81 @@
...
+        Any: Return value depends on the function called.
+
+    """
+    if not jinja_mappings_dict:
+        jinja_mappings_dict = _JINJA2_FUNCTION_MAPPINGS
+
+    result = {}
+
+    for function_name, function_import_path in jinja_mappings_dict.items():
+        if isinstance(function_import_path, dict):
```
itdependsnetworks, line 70: I don't think you need this, just spell it out in the mapping.

docs/source/netutils/utilities/index.rst
```diff
@@ -0,0 +1,65 @@
+*******************
+Netutils Utilities
+*******************
+
+Netutils to Jinja2 Filters
+============================
+
+In an effort to simplify the process of adding netutil's functions to jinja2 as filters we have created a convenience function. Let me give you an example of one use case.
```
jeffkala, line 8:
```suggestion
In an effort to simplify the process of adding netutils' functions to jinja2 as filters we have created a convenience function. One use case.
```

docs/source/netutils/utilities/index.rst
```diff
@@ -0,0 +1,65 @@
...
+    IP Address + 200 = 192.168.0.201
+
+
+Netutils to Jinja2 Filters List
+======================================
+
+The below list shows what jinja2 filters are added when you add them using the process above. The keys of the dictionary are the names you would use to call the jinja2 filter.
+
+.. note::
+    Some names have been changed to from their function names to provide more context when used as jinja2 filters.
```
jeffkala, line 57:
```suggestion
    Some names have been changed from their python function names to provide more context when used as a jinja2 filter.
```

netutils/utils.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def jinja2_convenience_function(jinja_mappings_dict=None, parent_folder=None):
+    """Convenience function that allows netutils filter to be used easily with jinja2.
+
+    Returns:
+        Any: Return value depends on the function called.
+
+    """
+    if not jinja_mappings_dict:
```
jeffkala, line 64: We want the ability to allow someone to specify their own mapping?  Not sure I understand why?

jeffkala (changes requested):

docs/source/netutils/utilities/index.rst
qduk: That does indeed work! Makes sense.

netutils/utils.py
qduk: I definitely made this more complicated than it needed to be. I blame it being Sunday when I did it and not know about `rstrip()`

netutils/utils.py
qduk: This has been refactored.

qduk:
@jeffkala @itdependsnetworks I refactored this. I don't know why I originally made it so complicated. I blame it being Sunday when I did this and not knowing about `rstrip()`.

netutils/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    "longest_prefix_match": "route.longest_prefix_match",
+    "vlanlist_to_config": "vlan.vlanlist_to_config",
+    "vlanconfig_to_list": "vlan.vlanconfig_to_list",
+}
+
+
+def jinja2_convenience_function():
+    """Convenience function that allows netutils filter to be used easily with jinja2.
+
+    Returns:
```
itdependsnetworks, line 58: Returns a dict, no?

itdependsnetworks: Can you add a testable example as well?

docs/source/netutils/utilities/index.rst
```diff
@@ -0,0 +1,64 @@
...
+Netutils to Jinja2 Filters
+============================
+
+In an effort to simplify the process of adding netutils' functions to jinja2 as filters we have created a convenience function. One use case.
+Here is the current folder structure.
+
+.. code-block:: python
+
+    .
+    ├── p1.py
```
itdependsnetworks, line 14: Use logical names

qduk: Updated

PR #53, opened 2021-09-24
Add get peer address [WIP]

itdependsnetworks:
There is no error handling, but this seems to work

### DIFF
--- a/netutils/ip.py
+++ b/netutils/ip.py
@@ -247,6 +247,35 @@ def get_first_usable(ip_network):
     return str(net[1])
 
 
+def get_peer_ip(ip_interface):
+    """Given an IP interface (an ip address, with subnet mask) that is on a peer network, return the peer IP.
+
+    Args:
+        ip_interface (str): An IP interface in string format that is able to be converted by `ipaddress` library.
+
+    Returns:
+        str: IP address formatted string with the cooresponding peer IP.
+
+    Example:
+        >>> from netutils.ip import get_peer_ip
+        >>> get_peer_ip('10.0.0.1/255.255.255.252')
+        '10.0.0.2'
+        >>> get_peer_ip('10.0.0.2/30')
+        '10.0.0.1'
+        >>> get_peer_ip('10.0.0.1/255.255.255.254')
+        '10.0.0.0'
+        >>> get_peer_ip('10.0.0.0/31')
+        '10.0.0.1'
+        >>>
+    """
+    ip_obj = ipaddress.ip_interface(ip_interface)
+    last_bit = int(ip_obj) & 1
+    mask_type = ip_obj._prefixlen & 1
+    if mask_type ^ last_bit:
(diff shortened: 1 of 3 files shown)


jeffkala (commented):
I know you said there is no error handling, but are we going to add it?  Think at a mininium we should validate the mask is correct?  Otherwise you can put any mask in and it just adds 
```
>>> get_peer_ip("10.0.0.1/32")
'10.0.0.2'
>>> get_peer_ip("10.0.0.1/24")
'10.0.0.2'
```

grelleum:
From the code snippet you have, it appears you are looking to get "the other" host address on a /31 or /30 network.
The following code will do that and I think it is a little more clear what it is doing, specifically it is finding the lowest numbered, non-conflicting host address on the same network as the ip_interface, so it will work with any prefix length that is not /32.
```
def get_peer(ip_interface):
    ip_obj = ipaddress.ip_interface(ip_interface)
    for host in ip_obj.network.hosts():
        if host == ip_obj.ip:
            continue
        ip_str = f"{host}/{ip_obj.netmask}"
        return ipaddress.ip_interface(ip_str)
```
This will return an IPv[46]Interface object.  To return an IPv[46]Address, simply `return host` after `continue`.

```
>>> repr(get_peer('10.20.30.1/32'))
'None'
>>> 
>>> get_peer('10.20.30.1/31')
IPv4Interface('10.20.30.0/31')
>>> 
>>> get_peer('10.20.30.1/30')
IPv4Interface('10.20.30.2/30')
>>> 
>>> get_peer('10.20.30.1/28')
IPv4Interface('10.20.30.2/28')
```

itdependsnetworks:
@jeffkala I should have said, no error handling "yet". 
@grelleum I am only considering /30, /31, /126, /128. Since there is already addition and subtraction, and it's the only place you can guarantee it will do the job you intend it to. Based on other's feedback, likely simplifying to be less efficient but more readable. 

This was suggested also, but again, not as readable. 
```
def get_peer_ip(ip_interface):
    ip_obj = ipaddress.ip_interface(ip_interface)
    mask = ((not(ip_obj._prefixlen & 1)) << 1) | 1
    return str(ipaddress.ip_address(int(ip_obj) ^ mask))
```

netutils/ip.py
```diff
@@ -247,6 +247,44 @@ def get_first_usable(ip_network):
...
 
 
+def get_peer_ip(ip_interface):
+    """Given an IP interface (an ip address, with subnet mask) that is on a peer network, return the peer IP.
+
+    Args:
+        ip_interface (str): An IP interface in string format that is able to be converted by `ipaddress` library.
+
+    Returns:
+        str: IP address formatted string with the cooresponding peer IP.
```
qduk, line 257:
```suggestion
        str: IP address formatted string with the corresponding peer IP.
```

qduk (approved):
What you have makes sense to me.

jeffkala (approved):
Look good Ken.

PR #54, opened 2021-09-27
Issue 1 duplicate line

grelleum:
Resolves duplicate line issue: https://github.com/networktocode/netutils/issues/1

Within IOSConfigParser._update_same_line_children_configs, a conditional has been added to see if a duplicated line is indented, and therefore part of a section, before assigning a parent.

### DIFF
--- a/netutils/__init__.py
+++ b/netutils/__init__.py
@@ -1,3 +1,3 @@
 """Initialization file for library."""
 
-__version__ = "0.2.2"
+__version__ = "0.2.3"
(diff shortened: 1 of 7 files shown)


netutils/__init__.py
```diff
@@ -1,3 +1,3 @@
 """Initialization file for library."""
 
-__version__ = "0.2.2"
+__version__ = "0.2.3"
```
itdependsnetworks, line 3: Thanks for pointing this out, this actually has to be fixed in a few places. You can remove this and squash, i'll update in another PR

jeffkala (approved):
I'm good with this after the other PR review comments are addressed.

itdependsnetworks:
Addressed in #128

Issue #56, opened 2021-09-30
Bandwidth Conversion


qduk:
### Environment
* netutils version:   0.2.3

### Proposed Functionality
Function for converting bandwidth speeds between kb, mb, gb etc

### Use Case
Converting between various speeds may be useful as certain APIs may give speed in different measurements.

dgarros:
For a project I'm working on right now i could use some functions to convert
- short bandwidth to an integer in kbps or bps , for example "100Gbps" to 100000000000 or "10Mbps" to 10000000
- kbps or bps to short bandwidth : 100000000000 to 100Gbps or 10000000 to 10Mbps

jdrew82:
> For a project I'm working on right now i could use some functions to convert
> 
> * short bandwidth to an integer in kbps or bps , for example "100Gbps" to 100000000000 or "10Mbps" to 10000000
> * kbps or bps to short bandwidth : 100000000000 to 100Gbps or 10000000 to 10Mbps

So you mean you'd like to be able to pass the speed name and have it interpret and convert it?

dgarros:
> So you mean you'd like to be able to pass the speed name and have it interpret and convert it?

Yes sir

PR #59, opened 2021-09-30
Add Bandwidth

jdrew82:
This PR includes methods for converting between Kbps, Mbps, Gbps, and Tbps as mentioned in #56. Tests are included and are passing locally.

### DIFF
--- a/netutils/bandwidth.py
+++ b/netutils/bandwidth.py
@@ -1,0 +1,146 @@
+"""Functions for performing bandwidth calculations."""
+import math
+
+
+def convert_Kbps_to_Mbps(speed: float) -> float:
+    """Method to convert `speed` from Kbps to Mbps value.
+
+    Args:
+        init_speed (float): Speed in Kbps to convert to Mbps.
+
+    Returns:
+        float: Speed in Mbps converted from Kbps.
+    """
+    return speed / 1000
+
+
+def convert_Kbps_to_Gbps(speed: float) -> float:
+    """Method to convert `speed` from Kbps to Gbps value.
+
+    Args:
+        init_speed (float): Speed in Kbps to convert to Gbps.
+
+    Returns:
+        float: Speed in Gbps converted from Kbps.
+    """
+    return speed / 1000000
+
+
+def convert_Kbps_to_Tbps(speed: float) -> float:
+    """Method to convert `speed` from Kbps to Tbps value.
+
+    Args:
+        init_speed (float): Speed in Kbps to convert to Tbps.
+
+    Returns:
+        float: Speed in Tbps converted from Kbps.
+    """
+    return speed / 1000000000
+
+
+def convert_Mbps_to_Gbps(speed: float) -> float:
+    """Method to convert `speed` from Mbps to Gbps value.
+
+    Args:
+        init_speed (float): Speed in Mbps to convert to Gbps.
+
+    Returns:
+        float: Speed in Mbps converted from Kbps.
+    """
+    return speed / 1000
+
+
+def convert_Mbps_to_Tbps(speed: float) -> float:
+    """Method to convert `speed` from Mbps to Tbps value.
+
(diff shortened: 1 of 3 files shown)


itdependsnetworks:
I was thinking of more generic functions. 

def bits_to_name():
def name_to_bits():
def bytes_to_name():
def name_to_bytes():

jdrew82:
> I was thinking of more generic functions.
> 
> def bits_to_name(): def name_to_bits(): def bytes_to_name(): def name_to_bytes():

I've got two methods added that will take a string and convert it to the Kbps and bps int form. Those should cover the `name_to_bits()` method. I'm unsure how to do a `bits_to_name()` method though as what is the name output we'd expect, Mbps? Gbps? As for the bytes methods, I figured bytes weren't necessary as bytes tend to relate to storage and less bandwidth/throughput. If we think it's worth doing though I can certainly work on adding them.

dgarros:
>  I'm unsure how to do a bits_to_name() method though as what is the name output we'd expect, Mbps? Gbps?
Yes Mbps? Gbps? sounds good to me, I would go with the closest match. If it is more than 1Gbps and less than 1Tbps go with Gbps, etc ... 

Thanks for implementing these functions so quickly, much appreciated

jdrew82:
@itdependsnetworks got your byte methods added in and the method names to match what you specified. Any others we should add?

netutils/bandwidth.py
```diff
@@ -0,0 +1,527 @@
...
+    """Method to convert a short bandwidth name to int value in bps.
+
+    Args:
+        speed (str): Bandwidth to be converted like `100Gbps` to bps.
+
+    Returns:
+        int: int value of bandwidth to be converted to bps
+    """
+    if re.search("[mM]bps", speed):
+        _value = name_to_kbits(speed) * 1000
```
dgarros, line 415: Looks like this like is the same as below, we should be able to combine them into a single if statement

netutils/bandwidth.py
```diff
@@ -0,0 +1,527 @@
...
+    """Method to convert a short bandwidth name to int value in KBps.
+
+    Args:
+        speed (str): Bandwidth to be converted like `100GBps` to KBps.
+
+    Returns:
+        int: int value of bandwidth to be converted to KBps
+    """
+    if re.search("[mM]Bps", speed):
+        _value = int(mbps_to_kbytes(int(re.sub("[mM]Bps", "", speed))))
```
dgarros, line 431: as far as. understand, this function will not be able to process `2.5Gbps` right now because we are only looking for an int

Since we are supporting decimals in `bits_to_name`, I think it would make sense to support decimal in all functions as well. it might be as simple as replacing all `int` with `float`

dgarros (commented):
thank you again for doing this

netutils/bandwidth.py
jdrew82: That's actually wrong. I have the wrong multiplier for Gbps I believe. I've corrected this.

netutils/bandwidth.py
```diff
@@ -0,0 +1,527 @@
...
+
+    Args:
+        speed (str): Bandwidth to be converted like `100GBps` to Bps.
+
+    Returns:
+        int: int value of bandwidth to be converted to Bps
+    """
+    if re.search("[mM]Bps", speed):
+        _value = int(mbps_to_kbytes(int(re.sub("[mM]Bps", "", speed))))
+    if re.search("[gG]Bps", speed):
```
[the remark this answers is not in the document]
jdrew82: Added

netutils/bandwidth.py
jdrew82: I've updated to use float and adjust the tests. The biggest problem I see now is that all answers will have a decimal point even if it's a whole number. If we're OK with that I'll leave it as is.

Issue #61, opened 2021-10-07
IndexError: list index out of range


jdrew82:
### Environment
* Python version:  3.7.9
* netutils version:  1.1.3

### Expected Behavior
Configuration is parsed without IndexError being thrown.

### Observed Behavior
Run a ConfigCompliance job from `nautobot-plugin-golden-config` and get the following error:

```
celery_1    | 15:44:21.188 ERROR   nautobot.jobs.ComplianceJob :
celery_1    |   RENDER COMPLIANCE TASK GROUP failed: list index out of range
celery_1    | [2021-10-06 15:44:21,188: ERROR/ForkPoolWorker-1] RENDER COMPLIANCE TASK GROUP failed: list index out of range
celery_1    | [2021-10-06 15:44:21,188: ERROR/ForkPoolWorker-1] example_host | RENDER COMPLIANCE TASK GROUP failed: list index out of range
celery_1    | [2021-10-06 15:44:21,203: ERROR/ForkPoolWorker-1] Host 'example_host': task 'RENDER COMPLIANCE TASK GROUP' failed with traceback:
celery_1    | Traceback (most recent call last):
celery_1    |   File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 589, in update_or_create
celery_1    |     obj = self.select_for_update().get(**kwargs)
celery_1    |   File "/usr/local/lib/python3.7/site-packages/cacheops/query.py", line 353, in get
celery_1    |     return qs._no_monkey.get(qs, *args, **kwargs)
celery_1    |   File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 431, in get
celery_1    |     self.model._meta.object_name
celery_1    | nautobot_golden_config.models.ConfigCompliance.DoesNotExist: ConfigCompliance matching query does not exist.
celery_1    |
celery_1    | During handling of the above exception, another exception occurred:
celery_1    |
celery_1    | Traceback (most recent call last):
celery_1    |   File "/usr/local/lib/python3.7/site-packages/nornir/core/task.py", line 99, in start
celery_1    |     r = self.task(self, **self.params)
celery_1    |   File "/usr/local/lib/python3.7/site-packages/nautobot_golden_config/nornir_plays/config_compliance.py", line 116, in run_compliance
celery_1    |     "extra": "",
celery_1    |   File "/usr/local/lib/python3.7/site-packages/django/db/models/manager.py", line 85, in manager_method
[...]

PR #62, opened 2021-10-07
Fix IndexError

jdrew82:
This PR should fix the issue in #61. A simple check of the length of the list before entering the loop should prevent an out of range error from being thrown.


jmcgill298 (commented):
This could actually give false information. I believe the issue is with duplicate lines in the root level config, and we probably need to update this logic to handle that.

jmcgill298:
It seems that I cannot undo my approval, but I think we should be cautious about the implications of this. I know that this would ignore duplicate comments for unnamed ACLs and some telephony commands

PR #66, opened 2021-10-28
Add failing test for IOS banner

pke11y:
This PR adds a failing test for an IOS banner.

It's objective is to draw attention to an issue parsing an IOS banner containing the `^` character as a banner delimiter.

The backup configuration is intentionally different to the intended to replicate the characters the Cisco IOS platform inserts when configuration is pasted into the IOS terminal.

### DIFF
--- a/tests/unit/mock/config/compliance/compliance/cisco_ios/ios_basic_backup.txt
+++ b/tests/unit/mock/config/compliance/compliance/cisco_ios/ios_basic_backup.txt
@@ -10,8 +10,8 @@ access-list 1 permit 10.10.20.20
 ntp server 192.168.0.100
 ntp server 192.168.0.101
 !
-banner login ^C
+banner login ^
 ******************
     TEST BANNER
 ******************
-^C
+^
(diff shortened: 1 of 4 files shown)


pke11y:
It would be good to get a bit of feedback on this issue. A customer has a long standing issue on the Nautobot GC plugin whereby configuration compliance cannot be completed due to the parsers inability to parse the banner section.
In this use case, the banner section is not a configured feature.

itdependsnetworks:
@jmcgill298 any objections to adding `^` as a valid delimiter?

jmcgill298:
My concern would be not capturing the following `C` character in the config on the device. I have used `\x03` to render this character in the config in the past. The code does convert that char and handle the diffs as well as handling additional `C` chars from people copying and pasting.

https://github.com/networktocode/netutils/blob/2a9690d4401ded5b7be66f46205ca27ce5130510/netutils/config/parser.py#L507-L508

I am not opposed to making this more flexible, I just want to make sure we don't cause another issue with the change.

itdependsnetworks:
Handled in #69

PR #75, opened 2021-11-17
Update Python Requirements

itdependsnetworks:


pyproject.toml
```diff
@@ -26,7 +27,7 @@ include = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.6.2"
+python = "^3.6"
```
jeffkala, line 30: Does this no longer break `black`?

.github/workflows/ci.yml
```diff
@@ -137,7 +137,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
```
jeffkala, line 140: 3.10 is interesting, trying to find the article but read somewhere that 3.10 and `black` don't interoperate accurately right now.

itdependsnetworks: tests passed :)

pyproject.toml
itdependsnetworks: Black is a development dependency, even is this is the case, it should not matter.

PR #76, opened 2021-11-20
Expose ipaddress for jinja filters

itdependsnetworks:


qduk:
@itdependsnetworks Do you want me to add a page in docs for this? Just explaining that it provides jinja2 filter for the python ipaddress library?

itdependsnetworks:
that would be great.

jeffkala (approved):

qduk:
@itdependsnetworks I didn't add a new page since we are just updating the `ip` with new functions but I did add a blurp in the utilities docs page.

Issue #78, opened 2021-11-23
F5 parser with iRules are in the configurations can capture extra sections


jeffkala:
### Environment
* Python version:  3.6
* netutils version:  develop branch

### Expected Behavior
Sections are parsed properly

### Observed Behavior
When iRules exist in the configuration and those iRules have when clauses and or tcl in them it throughs off the parsing logic and extra sections are found, this causes compliance to fail.

### Steps to Reproduce
1. Run the parser with an `ltm rule` parent that has tcl and comments within the block.

PR #79, opened 2021-11-24
F5 parser fix for irules with multiline single command lines.

jeffkala:
Add a preprocessing step that sanitizes the F5 configs completely removing the `ltm rule` configuration sections.  These sections are not uniform or standardized which makes a regex find or other solution very challenging.  ltm rule compliance checking is probably not frequently needed.

### DIFF
--- a/netutils/config/parser.py
+++ b/netutils/config/parser.py
@@ -625,7 +625,30 @@ class F5ConfigParser(BaseBraceConfigParser):
         return final_config
 
     def build_config_relationship(self):
-        r"""Parse text tree of config lines and their parents."""
+        r"""Parse text tree of config lines and their parents.
+        
+        Example:
+            >>> config = '''apm resource webtop-link aShare {
+            ...     application-uri http://funshare.example.com
+            ...     customization-group a_customization_group
+            ... }
+            ... apm sso form-based portal_ext_sso_form_based {
+            ...     form-action /Citrix/Example/ExplicitAuth/LoginAttempt
+            ...     form-field "LoginBtn Log+On
+            ... StateContext "
+            ...     form-password password
+            ...     form-username username
+            ...     passthru true
+            ...     start-uri /Citrix/Example/ExplicitAuth/Login*
+            ...     success-match-type cookie
+            ...     success-match-value CtxsAuthId
+            ... }
+            ... '''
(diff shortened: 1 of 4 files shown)


jeffkala:
fixes #78

netutils/config/parser.py
```diff
@@ -597,6 +597,104 @@ class F5ConfigParser(BaseBraceConfigParser):
...
+    def __init__(self, config):
+        """Create ConfigParser Object.
+
+        Args:
+            config (str): The config text to parse.
+        """
+        super(F5ConfigParser, self).__init__(self._clean_config_f5(config))
+
+    def _clean_config_f5(self, config_text):  # pylint: disable=no-self-use
+        """Removes all configuration items with 'ltm rule'.  iRules are essentially impossible to parse with the lack of uniformity.
```
jmcgill298, line 609:
```suggestion
        """Removes all configuration items with 'ltm rule'.
        
        iRules are essentially impossible to parse with the lack of uniformity,
        therefore, this method ensures they are not included in ``self.config``.
```

jmcgill298 (changes requested):

netutils/config/parser.py
```diff
@@ -597,6 +597,130 @@ class F5ConfigParser(BaseBraceConfigParser):
...
+
+        iRules are essentially impossible to parse with the lack of uniformity,
+        therefore, this method ensures they are not included in ``self.config``.
+
+        Args:
+            config_text (str): The entire config as a string.
+
+        Returns:
+            str: The sanitized config with all iRules (ltm rule) stanzas removed.
+        """
```
qduk, line 619: Should this function have an example? I don't know that we've defined is private functions should have an example. I just noticed this one doesn't and the one below does.

itdependsnetworks: I'll leave it to Jeff to decide, most private methods don't have it, but it is at times helpful to understand what is expected.

itdependsnetworks (approved):

netutils/config/parser.py
jeffkala: I don't think it makes sense.  In most cases I wouldn't expect someone to directly interact with a private method, by putting an example on it and having it in the docs might influence that.

PR #89, opened 2021-12-27
Added ip_to_ptr function

qduk:


netutils/ip.py
```diff
@@ -387,3 +387,23 @@ def get_usable_range(ip_network):
         lower_bound = str(net[1])
         upper_bound = str(net[-2])
     return f"{lower_bound} - {upper_bound}"
+
+
+def ip_to_ptr(ip_address):
+    """Generates a ptr record from a given IPv4 address.
```
itdependsnetworks, line 393: Update to inclued ipv6.

qduk: Updated

netutils/ip.py
```diff
@@ -387,3 +387,23 @@ def get_usable_range(ip_network):
         lower_bound = str(net[1])
         upper_bound = str(net[-2])
     return f"{lower_bound} - {upper_bound}"
+
+
+def ip_to_ptr(ip_address):
+    """Generates a ptr record from a given IP address.
+
+    Args:
+        ip4_address (str): IPv4 address.
```
jeffkala, line 396: Can we update this?

jeffkala (changes requested):

itdependsnetworks:
I am rethinking this one, I didn't realize that the ptr records already happen with Python ip address, the only one similar is broadcast, and perhaps that was not a great idea to begin with. You can use `netutils.ip.ipaddress_address` method to get the same thing. 

@dgarros what do you think?

netutils/ip.py
qduk: I should've caught that. Thanks Jeff. Updated in case we do use this PR.

qduk:
Closed per comment from @itdependsnetworks

PR #91, opened 2022-01-03
Dynamically gather start and end lines for doc testing

qduk:
Now, instead of defining a start and end line in both the `index.rst` file and in testing, testing pulls form the `index.rst` file and just ensures that the `start_value` and `end_value` are what you expect.

### DIFF
--- a/tests/unit/test_docs.py
+++ b/tests/unit/test_docs.py
@@ -3,6 +3,7 @@ import subprocess
 import glob
 import os
 import pytest
+import regex
 
 UNDOCUMENTED_FILES = ["__init__", "constants", "lib_mapper", "protocol_mapper", "variables"]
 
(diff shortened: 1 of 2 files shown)


tests/unit/test_docs.py
```diff
@@ -87,5 +113,7 @@ def test_folders_contain_index(data):
 
 @pytest.mark.parametrize("start_end", START_END_LINES, ids=[section["name"] for section in START_END_LINES])
 def test_docs_start_end_lines(start_end):
-    assert README_LIST[start_end["start_line"]] == start_end["start_value"]
-    assert README_LIST[start_end["end_line"]] == start_end["end_value"]
+    start_line_value = _get_start_line(start_end["name"])
+    end_line_value = _get_end_line(start_end["name"])
```
itdependsnetworks, line 117: Can you add one failure test as well? I feel like it hasn't always failed proactively.

qduk: I added a test to that asserts a different end value doesn't match the README. I'm not sure if that is sufficient though. I'm not sure how to test a failure as no exceptions get raised when it "fails"

tests/unit/test_docs.py
```diff
@@ -18,45 +19,70 @@
...
+            match = regex.search(start_line_regex, line)
+            if match:
+                break
+
+        if match:
+            int_value = int(match.groupdict()["start_value"])
+            return int_value
+
+        raise Exception(
+            f"Not able to parse end line value from {SPHINX_DIRECTORIES[0]['source_dir']}/{folder_name}/index.rst."
```
jeffkala, line 66:
```suggestion
            f"Not able to parse start line value from {SPHINX_DIRECTORIES[0]['source_dir']}/{folder_name}/index.rst."
```

tests/unit/test_docs.py
```diff
@@ -87,5 +113,7 @@ def test_folders_contain_index(data):
 
 @pytest.mark.parametrize("start_end", START_END_LINES, ids=[section["name"] for section in START_END_LINES])
 def test_docs_start_end_lines(start_end):
-    assert README_LIST[start_end["start_line"]] == start_end["start_value"]
-    assert README_LIST[start_end["end_line"]] == start_end["end_value"]
+    start_line_value = _get_start_line(start_end["name"])
+    end_line_value = _get_end_line(start_end["name"])
+    assert README_LIST[start_line_value] == start_end["start_value"]
+    assert README_LIST[end_line_value - 1] == start_end["end_value"]
```
jeffkala, line 119: Any reason we can't make this more try?  Perhaps by just passing in a arg of "start" or "end"?  Then have some constants for the regex patterns?

jeffkala (changes requested):

tests/unit/test_docs.py
qduk: This has been updated"

tests/unit/test_docs.py
qduk: After looking at this a little I'm not sure I know what you mean. Let's meet tomorrow to discuss quickly.

jeffkala (approved):

PR #97, opened 2022-02-08
Closes #92 Junos List

### LINKED ISSUE #92: Junos List Jinja Filter
### Environment
* netutils version:  * any future version

### Proposed Functionality
Junos requires lists to be formatted inline and within square brackets that are spaced. For example `authentication-order [ password radius ];` I propose a function to take a list and output a string that formats the list correctly. If there is only one item in the list the data is represented in the following format: `authentication-order password;`. 

### Use Case
Lets day we are doing `Network as Code` using Ansible and have defined the authentication order as the following:
```yaml
authentication_order:
  - password
  - radius
```

it would be nice to be able to easily convert the list above to valid junos syntax:
```jinja
authentication-order {{ authentication_order | junos_list }};
```

output but also handling the single item edge case:
```
authentication-order [ password radius ];
```

minitriga:
Implements `junos_list` to convert a list of items to a junos list. 

There may be a better way to format the string than what I have done please suggest ideas if it needs improving.


jeffkala:
As I mentioned in the issue itself.  I'm not really sure this is something that requires a function.  It can be done via something like this which uses all native builtin methods.
```python
>>> vlans = ["VLAN1", "VLAN2", "VLAN3", "VLAN4"]
>>>
>>> print(f"[ {' '.join(vlans)} ]")
[ VLAN1 VLAN2 VLAN3 VLAN4 ]
```

minitriga:
@jeffkala Yes it was implemented as more as a shortcut to handle single/multiple values without altering the templates. But yes if not needed ill close the Issue and MR.

PR #98, opened 2022-02-09
Nokia SrOS Parser

jdrew82:
This PR includes changes for adding a Nokia SrOS parser. This requires that the configuration be presented in "info full-context" format as opposed to the indented MD-CLI format. This format is similar to Juniper's set format. I've included a test showing the parser working and I've added a note about the config format in the docs.

### DIFF
--- a/tests/unit/mock/config/parser/nokia_sros/sros_full_sent.txt
+++ b/tests/unit/mock/config/parser/nokia_sros/sros_full_sent.txt
@@ -1,0 +1,256 @@
+configure { }
+configure { card 1 }
+configure { card 1 card-type iom-1 }
+configure { card 1 mda 1 }
+configure { card 1 mda 1 mda-type me12-100gb-qsfp28 }
+configure { card 1 mda 2 }
+configure { card 1 fp 1 }
+configure { log }
+configure { log filter "1001" }
+configure { log filter "1001" named-entry "10" }
+configure { log filter "1001" named-entry "10" description "Collect only events of major severity or higher" }
+configure { log filter "1001" named-entry "10" action forward }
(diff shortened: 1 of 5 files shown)


jeffkala (approved):
LGTM

tests/unit/mock/config/parser/nokia_sros/sros_full_sent.txt
```diff
@@ -0,0 +1,256 @@
...
+configure { system management-interface snmp streaming admin-state enable }
+configure { system bluetooth }
+configure { system bluetooth advertising-timeout 30 }
+configure { system login-control }
+configure { system login-control ssh }
+configure { system login-control ssh inbound-max-sessions 30 }
+configure { system security }
+configure { system security aaa }
+configure { system security aaa local-profiles }
+configure { system security aaa local-profiles profile "administrative" }
```
itdependsnetworks, line 73: Why are things like this quoted? Is there potential for a space, and would that effect the parsing?

jdrew82: This appears to be the formating of SrOS and from what I've seen, it's always a single space around the curly brackets. I believe this is a collapsed, single-line form of the tab indented format that doesn't include the repeated configuration lines, ie the `configuration system security aaa` lines in this example.

itdependsnetworks:
@jdrew82 does it make sense to go with this type of configuration? The thing is the nested configuration is likely the standard.

jdrew82:
No, I think it probably makes more sense to have the tab indented form parsed since that does appear to be the standard. Will need to update the parser to handle that.

itdependsnetworks:
based on previous, will close.

Issue #85, opened 2021-12-08
bits_to_name `decimal` handling seems odd


glennmatthews:
### Environment
* Python version:  3.9.7
* netutils version:  1.0.0

### Expected Behavior

```python
>>> bits_to_name(1234)
'1Kbps'
>>> bits_to_name(1234, 1)
'1.2Kbps'
>>> bits_to_name(1234, 2)
'1.23Kbps'
>>> bits_to_name(1234, 3)
'1.234Kbps'
>>> bits_to_name(1234, 4)
'1.2340Kbps'
```

### Observed Behavior

```python
>>> bits_to_name(1234)    # expect 0 decimal places, instead get 1 decimal place but it's wrong!
'1.0Kbps'
>>> bits_to_name(1234, 1)
'1.2Kbps'
>>> bits_to_name(1234, 2)
'1.23Kbps'
>>> bits_to_name(1234, 3)
'1.234Kbps'
>>> bits_to_name(1234, 4)  # expect 4 decimal places, instead still get 3 since the fourth is zero
'1.234Kbps'
```

jdrew82:
I observed this during testing I think. It's something with the way that `round` works I think.

itdependsnetworks:
@jdrew82 then perhaps using round is not the correct strategy or using round requires some post processing?

jdrew82:
Actually, now that I think about it, I think the problem is that it's a `float`. Because it's a `float` the string rep of it includes the `.0` portion. We'd need to format it to remove it.

jdrew82:
The issue appears to be that it will show the .0 if you specify the decimal point in `round` to be 0. If you don't specify it at all it doesn't include it. I've got a PR coming in with a tweak to address this quirk.

PR #99, opened 2022-02-09
Fix decimal place in bits_to_name

jdrew82:
This PR addresses issue #85. There appears to be a quirk with the round method that it will show the number as a float if you specify 0 decimal places and as an int, if you don't specify the places at all. This tweak should address the quirk and represent the name as expected.

### DIFF
--- a/netutils/bandwidth.py
+++ b/netutils/bandwidth.py
@@ -113,19 +113,20 @@ def bits_to_name(  # pylint: disable=too-many-branches,too-many-return-statement
     Example:
         >>> from netutils.bandwidth import bits_to_name
         >>> bits_to_name(125000)
-        '125.0Kbps'
+        '125Kbps'
         >>> bits_to_name(1000000000)
-        '1.0Gbps'
+        '1Gbps'
     """
     if not isinstance(speed, int):
         raise ValueError(f"Speed of {speed} was not a valid speed integer.")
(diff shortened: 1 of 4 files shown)


jeffkala (approved):

jeffkala (approved):
Looks good.

netutils/bandwidth.py
```diff
@@ -113,19 +113,23 @@ def bits_to_name(  # pylint: disable=too-many-branches,too-many-return-statement
...
+        '1Gbps'
     """
     if not isinstance(speed, int):
         raise ValueError(f"Speed of {speed} was not a valid speed integer.")
 
     for bit_type, val in BITS_MAPPING.items():
         if val["low"] <= speed < val["high"]:
             try:
-                return f"{round(speed / val['low'], nbr_decimal)}{bit_type}"
+                if nbr_decimal != 0:
```
itdependsnetworks, line 126: To keep it a bit more dry, does this make sense? 
```
if nbr_decimal == 0
    nbr_decimal = None
```

```
>>> round(1500 / 100, None)
15
>>> round(1500 / 100, 0)
15.0
>>>
```

jdrew82: Good call. I've refactored this bit to remove duplication and simplify it.

PR #104, opened 2022-03-12
Optimize vlanconfig_to_list using builtin Regex methods

FragmentedPacket:
Mostly learning opportunity, but took a shot at optimizing the code using RegEx findall method to find both regular vlans, and condensed VLANs and then only have to iterate over the condensed vlans to expand it.

### DIFF
--- a/netutils/vlan.py
+++ b/netutils/vlan.py
@@ -73,20 +73,25 @@ def vlanconfig_to_list(vlan_config):
         [1025, 1069, 1070, 1071, 1072, 1114, 1173, 1174, 1175, 1176, 1177, 1178, 1179, 1180, 1181, 1501, 1502]
         >>>
     """
-    vlans = []
-    for line in vlan_config.splitlines():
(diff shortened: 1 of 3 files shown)


netutils/vlan.py
```diff
@@ -73,20 +73,25 @@ def vlanconfig_to_list(vlan_config):
...
+
+    vlan_ranges = re.findall(r"\d+-\d+", vlan_config)
+    for v_range in vlan_ranges:
+        first, second = v_range.split("-")
+        # Add one to first to prevent duplicates that already exist within vlans
+        vlans.extend(list(range(*[int(first) + 1, int(second)])))
+
     vlans = sorted(vlans)
-    if vlans[-1] > 4094:
+    if max(vlans) > 4094:
```
itdependsnetworks, line 95: This would seem less optimized. With `max` I would guess it would have to evaluate each number in the list wheras the previous just checked the last one.

qduk (approved):

PR #105, opened 2022-03-12
Add min_grouping_size to vlanlist_to_config method

FragmentedPacket:
Defaults min_grouping_size to 3, but leaves it configurable for users.
Add necessary tests


jeffkala (approved):
LGTM

qduk (approved):

netutils/vlan.py
```diff
@@ -6,13 +6,14 @@
...
 
-def vlanlist_to_config(vlan_list, first_line_len=48, other_line_len=44):
+def vlanlist_to_config(vlan_list, first_line_len=48, other_line_len=44, min_grouping_size=3):
     """Given a List of VLANs, build the IOS-like vlan list of configurations.
 
     Args:
         vlan_list (list): Unsorted list of vlan integers.
         first_line_len (int, optional): The maximum length of the line of the first element of within the return list. Defaults to 48.
         other_line_len (int, optional): The maximum length of the line of all other elements of within the return list. Defaults to 44.
+        min_grouping_size (int, optional): The minimum grouping size. Defaults to Cisco's minimum grouping size of 3.
```
itdependsnetworks, line 16:
```suggestion
        min_grouping_size (int, optional): The minimum consecutive VLANs to aggregate with a hyphen . Defaults to Cisco's minimum grouping size of 3.
```

netutils/vlan.py
```diff
@@ -21,29 +22,36 @@ def vlanlist_to_config(vlan_list, first_line_len=48, other_line_len=44):
         >>> from netutils.vlan import vlanlist_to_config
         >>> vlanlist_to_config([1, 2, 3, 5, 6, 1000, 1002, 1004, 1006, 1008, 1010, 1012, 1014, 1016, 1018])
         ['1-3,5,6,1000,1002,1004,1006,1008,1010,1012,1014', '1016,1018']
-        >>>
+        >>> vlanlist_to_config([1,3,5,6,100,101,102,103,104,105,107,109], min_grouping_size=2)
+        ['1,3,5-6,100-105,107,109']
     """
+    # 1 is the loneliest number. Fail if min_grouping_size is less than 2.
```
itdependsnetworks, line 28: Why is that? Would it not be feasible for a vendor to supply no grouping? I realize this is similar to `",".join(vlan_list)`, but it still makes sense to use the same function for all vendors.

FragmentedPacket: Ah yes. I agree. I didn't account to just return the list if they didn't want the actual grouping by setting 1. Recent commit fixes this.

PR #106, opened 2022-03-18
Nokia SrOS Parser

jdrew82:
Here's my second attempt at a parser for Nokia SrOS. This time, I'm using the tab indented format that is standard for their configuration backups. This uses the standard space-based parser but also uses the section banners as parents for the section stripping out the section title from the 'echo "System Configuration"` portion. This theoretically allows the end-user to simply put in "System Configuration" when wanting to compare sections and it should parse the full block.


netutils/config/parser.py
```diff
@@ -949,3 +949,77 @@ def _build_nested_config(self, line):
...
+
+    def __init__(self, config):
+        """Create ConfigParser Object.
+
+        Args:
+            config (str): The config text to parse.
+        """
+        super(NokiaConfigParser, self).__init__(config)
+
+    def is_exit_next(self, line):  # pylint: disable=no-self-use
```
itdependsnetworks, line 968: I do not see this being used, I'm not sure how this is called?

netutils/config/parser.py
```diff
@@ -949,3 +949,77 @@ def _build_nested_config(self, line):
...
+            line (str): A config line from the device.
+
+        Returns:
+            bool: True if line has 'exit' or 'exit all', else False.
+        """
+        if line.lstrip() == "exit" or line.lstrip() == "exit all":
+            return True
+        return False
+
+    def is_section_title(self, line):  # pylint: disable=no-self-use
```
itdependsnetworks, line 981: I think is_section_title and get_section_title should be private methods by convention, e.g. `_is_section_title`

netutils/config/parser.py
```diff
@@ -949,3 +949,77 @@ def _build_nested_config(self, line):
...
+        return False
+
+    @property
+    def config_lines_only(self):
+        """Remove spaces and comments from config lines.
+
+        Returns:
+            str: The non-space and non-comment lines from ``config``.
+        """
+        # Specific to fortinet to remove uncommon data patterns for use later in _build_nested_config.
```
itdependsnetworks, line 1015: Remover reference to fortinet.

tests/unit/mock/config/parser/nokia_sros/sros_full_sent.txt
```diff
@@ -0,0 +1,5562 @@
+exit all
```
itdependsnetworks, line 1: Does it make sense to include this in the example? I am a bit torn (if I understand this correctly), what is the expected result, but we should be purposeful in our decision.

itdependsnetworks:
Looking good, but we def need more config tests.

netutils/config/parser.py
jdrew82: Sorry, this is left over from an earlier iteration. I had originally had it ignoring the `exit` clauses but have instead left them in. I've removed the method as it's no longer used.

netutils/config/parser.py
jdrew82: Done.

netutils/config/parser.py
jdrew82: Updated to be private.

tests/unit/mock/config/parser/nokia_sros/sros_full_sent.txt
jdrew82: I was simply copying this example from @h4ndzdatm0ld TTP repo with the assumption this is a standard format for configurations.

h4ndzdatm0ld: @itdependsnetworks @jdrew82 I am happy to share a ContainerLab setup with a Nokia SROS device for the team to use for current and future testing or integrations, etc. If interested, please let me know. Also, the basic 'show config' (admin display-config) equivalent of SROS will always include `exit all` at the end.

h4ndzdatm0ld: Also, on the needing more configs to test, I could dig up some old configs but they will need some sanitizing.

itdependsnetworks: that would be great.

netutils/config/parser.py
```diff
@@ -949,3 +949,63 @@ def _build_nested_config(self, line):
                 self.indent_level = spaces
 
             self._update_config_lines(line)
+
+
+class NokiaConfigParser(BaseSpaceConfigParser):
+    """Nokia SrOS config parser."""
+
+    comment_chars = ["# ", "#--------------------------------------------------"]
```
jeffkala, line 957: This confuses me a bit, I would think just having `#` would catch both of these?

jeffkala (approved):
Looks good, thanks!

tests/unit/mock/config/compliance/compliance/nokia_sros/sros_basic_feature.py
```diff
@@ -0,0 +1,4 @@
+features = [
+    {"name": "System Configuration", "ordered": False, "section": ["System Configuration"]},
```
itdependsnetworks, line 2: Just making sure that this is the "right" level of parent that we have. I am not as familiar with sros configs, but in ios and ios-like configs it's reasonable to have a single parent such as "router bgp" be a parent, do we have the "right" level of parent here?

jdrew82: I'm not a SrOS expert either but when it was discussed with the client they stated that it made the most sense to use the banners as section delimiters/parents due to the structure of the configuration. For example, if you ignore the section banners then it becomes quite difficult to try and define the parent line for a section as there is quite a bit of duplication throughout the config, ie `system` is at the same level in multiple sections and you need to potentially look at the next line and indentation to determine the differing section. If someone has a better idea of how to handle it I'm welcome to suggestions.

itdependsnetworks: @pszulczewski you have any thoughts?

pszulczewski: SROS config structure relies on indentation, delimiting contexts by 4 spaces, similarly to python. [You can even use python tokenizer to parse out sros onto bits, tested! ;-) ]
There is one parent `system` and other contexts like `system security` are children, in this case `security` is a child of parent `system` parent.
Banners are separating some contexts (child contexts in this case) only for the convenience of users, system doesn't really care when you deploy configuration.

In this case when you develop config compliance it makes sense to keep these sections to have better granularity for features. In this case you can treat separately `system` as System General, `system security` as Security and so on.

This looks to me like the right approach for config compliance.

jdrew82: Thanks for your confirmation that I'm on the right track @pszulczewski!

jeffkala:
Approved this PR, the final test that was added was properly being sent in as post processed config for SROS.

netutils/config/parser.py
```diff
@@ -949,3 +949,63 @@ def _build_nested_config(self, line):
...
+        return False
+
+    def _get_section_title(self, line):  # pylint: disable=no-self-use
+        """Determine section title from banner.
+
+        Args:
+            line (str): A config line from the device that has been found to be a section title.
+
+        Returns:
+            str|bool: The section's title from the section banner, else False.
```
qduk, line 988: I never thought about adding the `|` to specift `or`. That's cool.

qduk (approved):

jeffkala (approved):

PR #107, opened 2022-03-28
Fix issue when backup or intended is empty

itdependsnetworks:
Prior to this, there was an issue in which if the backup config or the intended config was empty, it would return None, instead of True/False for compliance.


wvandeun:
Solves an issue we were facing, when the intended config doesn't match any lines that were defined in the feature.

jeffkala (approved):

PR #113, opened 2022-04-25
Updated lib mapping docs

qduk:
This PR takes the old way we have library mappings in the docs to a new prettier HTML table.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -13,11 +13,14 @@
 import os
 import sys
 import toml
+from jinja2 import Environment, FileSystemLoader
+
 
 sys.path.insert(0, os.path.abspath("../.."))
+from netutils import lib_mapper  # noqa: E402
+
 sys.path.append(os.path.abspath("sphinxext"))
 toml_dict = toml.load("../../pyproject.toml")
-
 
 # -- Project information -----------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


qduk:
Screenshot 2022-04-25 003936

docs/source/conf.py
```diff
@@ -63,3 +66,21 @@
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ["_static"]
+
+
+def build_mapping_tables(app):
+    """Build the library mappings tables."""
+    env = Environment(loader=FileSystemLoader("docs/source"))
+    template_file = env.get_template("table_template.j2")
+    for dict_name in lib_mapper._LIST_OF_MAP_DICTS:
```
itdependsnetworks, line 75: What about something like this instead of static definition? 

```
from netutils import lib_mapper
LIST_OF_MAP_DICTS = []

for key in dir(lib_mapper):
    if key.endswith("MAPPER_REVERSE") or key.endswith("_MAPPER") and not key.startswith("_"):
        LIST_OF_MAP_DICTS.append(key)
```

qduk: Totally forgot about `dir()`. I added a few more conditionals in there to account for 2 of the mappings we don't pull into the docs. Maybe we want them in there though? It's the `MAIN` and `NETMIKO` mapping dictionaries.

docs/source/conf.py
```diff
@@ -63,3 +66,31 @@
...
 html_static_path = ["_static"]
+
+
+def build_mapping_tables(app):
+    """Build the library mappings tables."""
+    env = Environment(loader=FileSystemLoader("docs/source"))
+    template_file = env.get_template("table_template.j2")
+
+    LIST_OF_MAP_DICTS = []
+    print(dir(lib_mapper))
```
itdependsnetworks, line 77: Remove the print statements?

PR #114, opened 2022-04-28
Pinned mistune for m2r2

qduk:
Pinned mistune and updated dir paths so it runs successfully in readthedocs.


qduk:
docs can be found [here](https://ntc-netutils.readthedocs.io/en/mistune_pin/)

PR #118, opened 2022-06-08
Add IOS-XE parser support based on IOS

matt852:
Both Cisco IOS and Cisco IOS-XE use the same modules for Napalm, Ansible, etc.  I'd like to add parsing support for IOS-XE devices, using the same parser used for IOS.


jeffkala:
@matt852 thanks for the PR, I think I'm missing some additional context on why this is required?  Was there a certain failure or traceback that required this?  Based on some of the tests that were added its to handle compliance on certain lines like `name-server` vs `name server`?

itdependsnetworks:
I don't think there is enough of a difference between the two to warrant maintaining the additional code. Sorry for the extra work :(

PR #123, opened 2022-06-24
Add type hints to the whole project and mypy testing setup and CI

jeffkala:
The code base was previously mixed, some stuff was type hinted other stuff wasn't.  Went ahead got everything up to date.  Quite a few ignores are in place for now but can be cleaned up over time.


itdependsnetworks:
Pretty cool stuff!! Will ping @Kircheneer to ask for an extra pair of eyes.

Kircheneer:
Brilliant stuff @jeffkala!

I have a couple of meta-level comments and will dive into the down and dirty in a moment.

- As per [this](https://google.github.io/styleguide/pyguide.html#doc-function-args) and DRY I believe we should drop type annotations from the docstrings wherever code-level type hints are present
- If we add an empty file called `py.typed` to the python module root other modules importing `netutils` can also benefit from its type hinting (see [here](https://peps.python.org/pep-0561/#packaging-type-information))

netutils/bandwidth.py
```diff
@@ -123,7 +124,7 @@ def bits_to_name(  # pylint: disable=too-many-branches,too-many-return-statement
     for bit_type, val in BITS_MAPPING.items():
         if val["low"] <= speed < val["high"]:
             if nbr_decimal == 0:
-                nbr_decimal = None
+                nbr_decimal = None  # type: ignore
```
Kircheneer, line 127: I believe that `round(x, None) == round(x)` for all `x`, right? Meaning we could drop this branch and the corresponding `# type: ignore`.

netutils/config/clean.py
```diff
@@ -1,11 +1,11 @@
 """Functions for working with configuration to clean the config."""
 # pylint: disable=anomalous-backslash-in-string
 
-
 import re
+import typing as t
 
 
-def clean_config(config, filters):
+def clean_config(config: str, filters: t.List[str]) -> str:
```
Kircheneer, line 8: The docstring example for this doesn't pass type checking. Now AFAIK there is no way for Mypy to perform type checking on the code in the examples, but I still think it's worth pointing out: `clean_filters` in that example is a `List[Dict[str, str]]` rather than the `List[str]` annotation in the function signature here.

netutils/config/clean.py
```diff
@@ -48,11 +48,11 @@ def clean_config(config, filters):
         >>>
     """
     for item in filters:
-        config = re.sub(item["regex"], "", config, flags=re.MULTILINE)
+        config = re.sub(item["regex"], "", config, flags=re.MULTILINE)  # type: ignore
```
Kircheneer, line 51: See last comment, can be removed if the function signature is fixed.

netutils/config/clean.py
```diff
@@ -75,6 +75,6 @@ def sanitize_config(config, filters):
         'enable secret 5 <removed>'
         >>>
     """
-    for item in filters:
-        config = re.sub(item["regex"], item["replace"], config, flags=re.MULTILINE)
+    for item in filters:  # type: ignore
+        config = re.sub(item["regex"], item["replace"], config, flags=re.MULTILINE)  # type: ignore
```
Kircheneer, line 79: Same thing as with `clean_config`

netutils/config/compliance.py
```diff
@@ -1,8 +1,10 @@
 """Filter Plugins for compliance checks."""
 
+import typing as t
+
 from . import parser  # pylint: disable=relative-beyond-top-level
 
-parser_map = {
+parser_map: t.Dict[str, t.Callable] = {  # type: ignore
```
Kircheneer, line 7: Type of `parser_map` should be `t.Dict[str, t.Type[BaseConfigParser]]`. `t.Type[SomeClass]` is the type hint for the actual class type or any subclasses (rather than instances of the class).

netutils/config/compliance.py
```diff
@@ -15,7 +17,7 @@
     "nokia_sros": parser.NokiaConfigParser,
 }
 
-default_feature = {
+default_feature: t.Dict[str, t.Union[str, bool, None]] = {
```
Kircheneer, line 20: Would it possibly make sense to use a [`TypedDict`](https://peps.python.org/pep-0589/) here? It seems like `default_feature` is somewhat struct-esque, with a `TypedDict` we could clearly identify the types of all its values.

netutils/config/compliance.py
```diff
@@ -27,7 +29,7 @@
 }
 
 
-def _check_configs_differences(intended_cfg, actual_cfg, network_os):
+def _check_configs_differences(intended_cfg: str, actual_cfg: str, network_os: str) -> t.Dict[str, t.Union[str, bool]]:
```
Kircheneer, line 32: If I see this correctly this would be another place we could use the proposed `default_feature` `TypedDict`.

netutils/config/compliance.py
```diff
@@ -97,17 +99,23 @@ def _is_feature_ordered_compliant(feature_intended_cfg, feature_actual_cfg):
...
-def _open_file_config(cfg_path):
+def _open_file_config(cfg_path: str) -> t.Union[str, bool]:
     """Open config file from local disk."""
     try:
         with open(cfg_path, encoding="utf-8") as filehandler:
             device_cfg = filehandler.read()
     except IOError:
-        return False
+        return False  # This should probably be changed to a exception raise. Causing mypy issues on L183, L184
     return device_cfg.strip()
```
Kircheneer, line 109: +1 on the exception part. The return type for this looks weird and would introduce lots of unnecessary checks because what the user of this function probably wants is the `str`. If something points to an exception not being correct here, how about an empty string?

netutils/config/compliance.py
```diff
@@ -97,17 +99,23 @@ def _is_feature_ordered_compliant(feature_intended_cfg, feature_actual_cfg):
...
 
 
-def compliance(features, backup, intended, network_os, cfg_type="file"):
+def compliance(
+    features: t.List[t.Dict[str, t.Union[str, bool, t.List[str]]]],
```
Kircheneer, line 113: Another possible place to use the `default_feature` `TypedDict`?

netutils/config/compliance.py
```diff
@@ -97,17 +99,23 @@ def _is_feature_ordered_compliant(feature_intended_cfg, feature_actual_cfg):
...
+    backup: str,
+    intended: str,
+    network_os: str,
+    cfg_type: t.Optional[str] = "file",
```
Kircheneer, line 117: If we supply a default argument for this but make the type `Optional`, do we expect users to explicitly pass `None` to `cfg_type` or can we just omit the `Optional`?

netutils/config/compliance.py
```diff
@@ -172,14 +180,16 @@ def compliance(features, backup, intended, network_os, cfg_type="file"):
     compliance_results = {}
 
     for feature in features:
-        backup_str = section_config(feature, backup_cfg, network_os)
-        intended_str = section_config(feature, intended_cfg, network_os)
+        backup_str = section_config(feature, backup_cfg, network_os)  # type: ignore
+        intended_str = section_config(feature, intended_cfg, network_os)  # type: ignore
```
Kircheneer, line 184: This could be fixed by changing the return type of `open_file_config` to just a `str`.

netutils/config/compliance.py
```diff
@@ -172,14 +180,16 @@ def compliance(features, backup, intended, network_os, cfg_type="file"):
...
         compliance_results.update({feature["name"]: feature_compliance(feature, backup_str, intended_str, network_os)})
 
-    return compliance_results
+    return compliance_results  # type: ignore
 
 
-def config_section_not_parsed(features, device_cfg, network_os):
+def config_section_not_parsed(
+    features: t.List[t.Dict[str, t.Union[str, bool, t.List[str]]]], device_cfg: str, network_os: str
```
Kircheneer, line 191: Do we want `network_os` to be an Enum type? This way we further restrict possible values for it and make sure we don't accidentally pass wrong types to it. Also having `netutils` possibly even export such an Enum for use within importing libraries makes conceptual sense to me.

netutils/config/compliance.py
```diff
@@ -418,7 +430,7 @@ def section_config(feature, device_cfg, network_os):
                 continue
             else:
                 match = False
-        for line_start in section_starts_with:
+        for line_start in section_starts_with:  # type: ignore
```
Kircheneer, line 433: Could be fixed with the `feature` `TypedDict` as `feature.get("section")` would return something whose type Mypy would be able to infer correctly.

netutils/config/parser.py
```diff
@@ -12,34 +13,34 @@
 class BaseConfigParser:  # pylint: disable=too-few-public-methods
     """Base class for parsers."""
 
-    comment_chars = ["!"]
-    banner_start = ["banner", "vacant-message"]
+    comment_chars: t.List[str] = ["!"]
+    banner_start: t.List[str] = ["banner", "vacant-message"]
```
Kircheneer, line 17: These type hints are not needed, Mypy is able to infer this type correctly. I think we should drop them, as then they just add noise to the code.

netutils/config/parser.py
```diff
@@ -12,34 +13,34 @@
...
+        self._config: t.Optional[str] = None
+        self._current_parents: t.Tuple[str, ...] = ()
+        self.generator_config: t.Generator[str, None, None] = (line for line in self.config_lines_only.splitlines())  # type: ignore
+        self.config_lines: t.List[ConfigLine] = []
+        self.build_config_relationship()  # type: ignore
 
-    def config_lines_only(self):
+    def config_lines_only(self):  # type: ignore
         """Remove lines not related to config."""
         raise NotImplementedError
```
Kircheneer, line 34: How about:
```python
class BaseConfigParser(metaclass=ABCMeta):
    ...
    def __init(self, config: str):
        ...
        self.build_config_relationship()

    @abstractmethod
    def config_lines_only(self):  # type: ignore
        """Remove lines not related to config."""
        pass

    @abstractmethod
    def build_config_relationship(self) -> t.List[ConfigLine]:
        """Parse text tree of config lines and their parents."""
        pass
```

netutils/config/parser.py
```diff
@@ -12,34 +13,34 @@
...
 
 
 class BaseSpaceConfigParser(BaseConfigParser):
     """Base parser class for config syntax that demarcates using spaces/indentation."""
 
-    comment_chars = ["!"]
-    banner_start = ["banner", "vacant-message"]
+    comment_chars: t.List[str] = ["!"]
+    banner_start: t.List[str] = ["banner", "vacant-message"]
```
Kircheneer, line 41: Can also get rid of these annotations.

netutils/config/parser.py
```diff
@@ -66,11 +67,11 @@ def is_banner_end(self, line):
         Returns:
             bool: True if line ends banner, else False.
         """
-        if self.banner_end in line:
+        if self.banner_end in line:  # type: ignore
```
Kircheneer, line 70: Would be fixed by the proposed changes to the base class.

netutils/config/parser.py
```diff
@@ -156,7 +157,7 @@ def get_leading_space_count(config_line):
         """
         return len(config_line) - len(config_line.lstrip())
 
-    def _remove_parents(self, line, current_spaces):
+    def _remove_parents(self, line: str, current_spaces: int) -> t.Tuple:  # type: ignore
```
Kircheneer, line 160: `t.Tuple[str, ...]` as the return type should be able to get rid of the `# type: ignore`. Correspons to a tuple of varying size with `str` elements.

netutils/config/parser.py
```diff
@@ -178,7 +179,7 @@ def _remove_parents(self, line, current_spaces):
         parents = self._current_parents[:-deindent_level] or (self._current_parents[0],)
         return parents
 
-    def _build_banner(self, config_line):
+    def _build_banner(self, config_line: str) -> t.Union[str, None]:
```
Kircheneer, line 182: We could use `t.Optional[str]` here to simplify the type a little

netutils/config/parser.py
```diff
@@ -198,7 +199,7 @@ def _build_banner(self, config_line):
             if not self.is_banner_end(line):
                 banner_config.append(line)
             else:
-                line = normalise_delimiter_caret_c(self.banner_end, line)
+                line = normalise_delimiter_caret_c(self.banner_end, line)  # type: ignore
```
Kircheneer, line 202: Fixed by proposed changes to the base class.

netutils/config/parser.py
```diff
@@ -213,7 +214,7 @@ def _build_banner(self, config_line):
 
         raise ValueError("Unable to parse banner end.")
 
-    def _build_nested_config(self, line):
+    def _build_nested_config(self, line: str) -> t.Optional[str]:  # type: ignore
```
Kircheneer, line 217: If we stick a `return None` at the end of the function the behavior of the function should stay the same but we can get rid of the `# type: ignore`.

netutils/config/parser.py
```diff
@@ -246,15 +247,15 @@ def _build_nested_config(self, line):
                 self.indent_level = spaces
 
             if self.is_banner_start(line):
-                line = self._build_banner(line)
+                line = self._build_banner(line)  # type: ignore
                 if line is None or not line[0].isspace():
                     self._current_parents = ()
                     self.indent_level = 0
                     return line
```
Kircheneer, line 254: We should be able to get rid of this `# type: ignore` by using a different variable name than the item of the for loop here:
```
            if self.is_banner_start(line):
                current_line = self._build_banner(line)
                if current_line is None or not current_line[0].isspace():
                    self._current_parents = ()
                    self.indent_level = 0
                    return current_line
```

netutils/config/parser.py
```diff
@@ -291,24 +292,24 @@ def build_config_relationship(self):
...
-                        line = self._build_nested_config(line)
+                        line = self._build_nested_config(line)  # type: ignore
                     else:
                         self._current_parents = ()
 
             if line is None:
                 break
             elif self.is_banner_start(line):
-                line = self._build_banner(line)
+                line = self._build_banner(line)  # type: ignore
```
Kircheneer, line 312: All these comments should be resolved by the proposed changes to the various methods.

netutils/config/parser.py
```diff
@@ -375,7 +376,7 @@ def build_config_relationship(self):
 
         return self.config_lines
 
-    def _build_multiline_config(self, delimiter):
+    def _build_multiline_config(self, delimiter: str):  # type: ignore
```
Kircheneer, line 379: Resolvable through:
- Adding `-> t.Optional[ConfigLine]` to the function signature
- Add an explicit default `return None` at the end
- Checking for None in all the places this function is called from

netutils/config/parser.py
```diff
@@ -461,20 +462,20 @@ def is_banner_one_line(config_line):
...
         if state:
             self.banner_end = line
         return state
 
     @property
-    def banner_end(self):
+    def banner_end(self) -> str:
         """Demarcate End of Banner char(s)."""
-        return self._banner_end
+        return self._banner_end  # type: ignore
```
Kircheneer, line 475: `self._banner_end` evaluates to `Optional[str]` if I remove this ignore comment, therefore we either
- perform explicit Optional handling in this property or
- return an `Optional[str]` from this

netutils/config/parser.py
```diff
@@ -484,17 +485,17 @@ def banner_end(self, banner_start_line):
...
             config (str): The config text to parse.
         """
-        self.unique_config_lines = set()
-        self.same_line_children = set()
+        self.unique_config_lines: t.Set[ConfigLine] = set()
+        self.same_line_children: t.Set[ConfigLine] = set()
         super(IOSConfigParser, self).__init__(config)
 
-    def _build_banner(self, config_line):
+    def _build_banner(self, config_line: str) -> t.Union[str, None]:
```
Kircheneer, line 498: Good place to use `t.Optional`

netutils/config/parser.py
```diff
@@ -572,17 +573,17 @@ class NXOSConfigParser(CiscoConfigParser, BaseSpaceConfigParser):
...
             config (str): The config text to parse.
         """
-        self.unique_config_lines = set()
-        self.same_line_children = set()
+        self.unique_config_lines: t.Set[ConfigLine] = set()
+        self.same_line_children: t.Set[ConfigLine] = set()
         super(NXOSConfigParser, self).__init__(config)
 
-    def _build_banner(self, config_line):
+    def _build_banner(self, config_line: str) -> t.Union[str, None]:
```
Kircheneer, line 586: `t.Optional`

netutils/config/parser.py
```diff
@@ -700,7 +701,7 @@ def build_config_relationship(self):
 
         return self.config_lines
 
-    def _build_multiline_single_configuration_line(self, delimiter, prev_line):
+    def _build_multiline_single_configuration_line(self, delimiter: str, prev_line: str):  # type: ignore
```
Kircheneer, line 704: Resolvable through:

- Adding -> t.Optional[ConfigLine] to the function signature
- Add an explicit default return None at the end
- Checking for None in all the places this function is called from

netutils/config/parser.py
```diff
@@ -820,7 +821,7 @@ def build_config_relationship(self):
                 self._current_parents = (previous_config.config_line,)
                 self.indent_level = self.get_leading_space_count(line)
                 if line is not None and line[0].isspace():
-                    line = self._build_nested_config(line)
+                    line = self._build_nested_config(line)  # type: ignore
```
Kircheneer, line 824: *Probably*  resolved by resolving the other ignore comments.

netutils/config/parser.py
```diff
@@ -914,7 +915,7 @@ def _get_uncommon_lines(self, config):  # pylint: disable=no-self-use
             result.update({group_match[0].split('"')[1]: group_match[1]})
         return result
 
-    def _build_nested_config(self, line):
+    def _build_nested_config(self, line: str) -> t.Optional[str]:  # type: ignore
```
Kircheneer, line 918: Just needs an explicit `return None` at the end of the function and we're good!

netutils/config/parser.py
```diff
@@ -928,7 +929,7 @@ def _build_nested_config(self, line):
             IndexError: When the number of parents does not match the expected deindent level.
         """
         if "[" in line:
-            line = self.uncommon_data.get(line.split('"')[1])
+            line = self.uncommon_data.get(line.split('"')[1])  # type: ignore
```
Kircheneer, line 932: Should perform a bit of explicit optional handling here to get rid of the ignore comment.

netutils/config/parser.py
```diff
@@ -978,7 +979,7 @@ def _is_section_title(self, line):  # pylint: disable=no-self-use
             return True
         return False
 
-    def _get_section_title(self, line):  # pylint: disable=no-self-use
+    def _get_section_title(self, line: str) -> t.Union[str, bool]:  # pylint: disable=no-self-use
```
Kircheneer, line 982: In the `False` case, should we return an empty string? This would be a somewhat breaking change, but the empty string is still fals-ey (`bool("") == False`).

netutils/interface.py
```diff
@@ -35,7 +36,7 @@ def _range_expand(regex_match):
                 number_range.append(int(value))
         return number_range
 
-    def _pairwise(interface_constant):
+    def _pairwise(interface_constant: t.List[int]) -> t.List[t.Iterable[t.Any]]:
         interface_constant_it = iter(interface_constant)
         return list(zip(interface_constant_it, interface_constant_it))
```
Kircheneer, line 41: If this really just wants to handle lists of integers, we can use the following signature:
```
def _pairwise(interface_constant: t.List[int]) -> t.List[t.Tuple[int, int]]:
```

If we need to handle arbitrary data types, which the `t.Any` in the return value suggests, we will need some [`TypeVar`](https://docs.python.org/3/library/typing.html#typing.TypeVar) usage here.

netutils/interface.py
```diff
@@ -57,14 +58,14 @@ def _pairwise(interface_constant):
     for element in itertools.product(*cartesian_list):
         current_interface = ""
         for count, item in enumerate(interface_constant_out):
-            current_interface += interface_pattern[item[0] : item[1]]  # noqa: E203
+            current_interface += interface_pattern[item[0] : item[1]]  # type: ignore # noqa: E203
```
Kircheneer, line 61: Resolved by resolving `_pairwise`.

netutils/interface.py
```diff
@@ -86,7 +87,9 @@ def split_interface(interface):
     return (head, tail)
 
 
-def canonical_interface_name(interface, addl_name_map=None, verify=False):
+def canonical_interface_name(
+    interface: str, addl_name_map: t.Optional[t.Dict[str, str]] = None, verify: t.Optional[bool] = False
```
Kircheneer, line 91: Does verify need the possibility of being `None`?

netutils/interface.py
```diff
@@ -178,15 +187,20 @@ def canonical_interface_name_list(interfaces, addl_name_map=None, verify=False,
         raise ValueError(f"Verify interface on and no match found for {no_match_string}")
 
     if order:
-        canonical_interface_list = INTERFACE_LIST_ORDERING_OPTIONS.get(order)(canonical_interface_list)
+        canonical_interface_list = INTERFACE_LIST_ORDERING_OPTIONS.get(order)(canonical_interface_list)  # type: ignore
```
Kircheneer, line 190: If we perform explicit None handling here we can get rid of the ignore comment.

netutils/interface.py
```diff
@@ -178,15 +187,20 @@ def canonical_interface_name_list(interfaces, addl_name_map=None, verify=False,
...
 
     return canonical_interface_list
 
 
-def abbreviated_interface_name(interface, addl_name_map=None, addl_reverse_map=None, verify=False):
+def abbreviated_interface_name(
+    interface: str,
+    addl_name_map: t.Optional[t.Dict[str, str]] = None,
+    addl_reverse_map: t.Optional[t.Dict[str, str]] = None,
+    verify: t.Optional[bool] = False,
```
Kircheneer, line 202: Does `verify` need the possibility of being `None`?

netutils/interface.py
```diff
@@ -237,7 +251,7 @@ def abbreviated_interface_name(interface, addl_name_map=None, addl_reverse_map=N
     return interface
 
 
-@total_ordering
+@total_ordering  # type: ignore
```
Kircheneer, line 254: We could add this issue to this comment: https://github.com/python/mypy/issues/5374

netutils/interface.py
```diff
@@ -280,8 +294,8 @@ def __hash__(self) -> int:  # noqa: D105
 class CCString(CharacterClass):
     """Strings are sorted lexicographically."""
 
-    def __lt__(self, other) -> bool:  # noqa: D105
-        return self.weight < other.weight or self.val < other.val
+    def __lt__(self, other) -> bool:  # type: ignore  # noqa: D105
```
Kircheneer, line 297: Can be fixed with `def __lt__(self, other: "CharacterClass") -> bool:  # noqa: D105`

netutils/interface.py
```diff
@@ -294,7 +308,7 @@ def weight(self) -> int:  # noqa: D107,D102
 class CCInt(CharacterClass):
     """Ints must be sorted canonically because '11' < '5'."""
 
-    def __lt__(self, other) -> bool:  # noqa: D105
+    def __lt__(self, other) -> bool:  # type: ignore  # noqa: D105
```
Kircheneer, line 311: Can be fixed with def __lt__(self, other: "CharacterClass") -> bool: # noqa: D105

netutils/interface.py
```diff
@@ -310,7 +324,7 @@ class CCSeparator(CharacterClass):
 
     weights: t.Dict[str, int] = {".": 10, "/": 20}
 
-    def __lt__(self, other) -> bool:  # noqa: D105
+    def __lt__(self, other) -> bool:  # type: ignore  # noqa: D105
```
Kircheneer, line 327: Can be fixed with `def __lt__(self, other: "CharacterClass") -> bool: # noqa: D105`

netutils/interface.py
```diff
@@ -345,15 +359,15 @@ def _split_interface_tuple(interface: str) -> t.Tuple[CharacterClass, ...]:
                 part += interface[idx]
                 idx += 1
             if part and idx == len(interface):
-                tail = (*tail, cls(part, True))
+                tail = (*tail, cls(part, True))  # type: ignore
                 break
             if part:
-                tail = (*tail, cls(part))
+                tail = (*tail, cls(part))  # type: ignore
```
Kircheneer, line 365: There probably is a way to resolve these, but I deem it not worth it for now. Something like `Union[Callable[[str, Optional[bool]], CharacterClass], Callable[[...], NoReturn]` might do the trick if you want to play around with it a little.

netutils/interface.py
```diff
@@ -500,15 +519,15 @@ def abbreviated_interface_name_list(  # pylint: disable=R0913, R0914
         raise ValueError(f"Verify interface on and no match found for {no_match_string}")
 
     if order:
-        abbreviated_interface_list = INTERFACE_LIST_ORDERING_OPTIONS.get(order)(abbreviated_interface_list)
+        abbreviated_interface_list = INTERFACE_LIST_ORDERING_OPTIONS.get(order)(abbreviated_interface_list)  # type: ignore
```
Kircheneer, line 522: Resolvable through explicit optional handling

netutils/ip.py
```diff
@@ -174,13 +176,13 @@ def is_ip(ip):
         >>>
     """
     try:
-        ip = ipaddress.ip_address(ip)
+        ip = ipaddress.ip_address(ip)  # type: ignore
```
Kircheneer, line 179: Just doing `ipaddress.ip_address(ip)` here should work to get rid of the comment.

netutils/ip.py
```diff
@@ -221,11 +223,11 @@ def netmask_to_cidr(netmask):
         23
     """
     if is_netmask(netmask):
-        return bin(int(ipaddress.ip_address(netmask))).count("1")
+        return bin(int(ipaddress.ip_address(netmask))).count("1")  # type: ignore
```
Kircheneer, line 226: Should change the return value of the function to `int`, then we can get rid of the ignore comment.

netutils/ip.py
```diff
@@ -269,7 +271,7 @@ def cidr_to_netmaskv6(cidr):
     raise ValueError("Parameter must be an integer between 0 and 128.")
 
 
-def get_all_host(ip_network):
+def get_all_host(ip_network: str) -> t.List[str]:
```
Kircheneer, line 274: `-> Generator[str, None, None]` and then we can also get rid of the ignore comment

netutils/ip.py
```diff
@@ -326,7 +328,7 @@ def get_first_usable(ip_network):
     return str(net[1])
 
 
-def get_peer_ip(ip_interface):
+def get_peer_ip(ip_interface: str) -> t.Any:
```
Kircheneer, line 331: Should be able to just return `str` here

netutils/mac.py
```diff
@@ -2,26 +2,27 @@
...
         if kwargs.get("mac"):
             mac = kwargs.get("mac")
         else:
             mac = args[0]
-        if not is_valid_mac(mac):
+        if not is_valid_mac(mac):  # type: ignore
             raise ValueError(f"There was not a valid mac address in: `{mac}`")
         return func(*args, **kwargs)
 
     return decorated
```
Kircheneer, line 22: Could still be more precise, but this should be enough for this:
```def _valid_mac(func: t.Callable[..., t.Any]) -> t.Callable[..., t.Any]:
    """Decorator to validate a MAC address is valid."""

    @wraps(func)
    def decorated(*args: t.Any, **kwargs: t.Any) -> t.Any:
        if kwargs.get("mac"):
            mac = kwargs.get("mac")
        else:
            mac = args[0]
        assert isinstance(mac, str)
        if not is_valid_mac(mac):
            raise ValueError(f"There was not a valid mac address in: `{mac}`")
        return func(*args, **kwargs)

    return decorated
```

netutils/mac.py
```diff
@@ -44,8 +45,8 @@ def is_valid_mac(mac):
     return False
 
 
-@_valid_mac
-def mac_to_format(mac, frmt="MAC_NO_SPECIAL"):
+@_valid_mac  # type: ignore
```
Kircheneer, line 48: Can get rid of all of these once the decorator is fixed

netutils/password.py
```diff
@@ -103,12 +106,14 @@ def compare_type5(unencrypted_password, encrypted_password, return_original=Fals
     salt = get_hash_salt(encrypted_password)
     if encrypt_type5(unencrypted_password, salt) == encrypted_password:
         if return_original is True:
-            return encrypted_password
+            return encrypted_password  # type: ignore
```
Kircheneer, line 109: Return value of the function needs to be `Union[str, bool]`, then we can get rid of this (but need to probably perform explicit handling of that return type in lots of places).

netutils/password.py
```diff
@@ -206,25 +211,25 @@ def encrypt_type7(unencrypted_password, salt=None):
...
     for i, _ in enumerate(unencrypted_password):
-        hex_password = "%02x" % (ord(unencrypted_password[i]) ^ XLAT[salt])  # pylint: disable=consider-using-f-string
+        hex_password = "%02x" % (ord(unencrypted_password[i]) ^ XLAT[salt])  # type: ignore # pylint: disable=consider-using-f-string
         encrypted_password += hex_password
-        salt += 1
-        if salt == 51:
-            salt = 0
+        salt += 1  # type: ignore
+        if salt == 51:  # type: ignore
+            salt = 0  # type: ignore
```
Kircheneer, line 221: If we use a different variable name for the changed form of `salt`, we can get by without all the ignore comments

netutils/ping.py
```diff
@@ -1,8 +1,9 @@
 """Functions to create a ping via pure Python."""
 import socket
+import typing as t
 
 
-def tcp_ping(ip, port, timeout=1):  # pylint: disable=invalid-name
+def tcp_ping(ip: str, port: int, timeout: t.Optional[int] = 1) -> bool:  # pylint: disable=invalid-name
```
Kircheneer, line 6: Can `timeout` be `None` in any case? If not, we can also get rid of the below ignore comments

netutils/vlan.py
```diff
@@ -1,12 +1,17 @@
...
 from itertools import groupby
+from operator import itemgetter
 
 
-def vlanlist_to_config(vlan_list, first_line_len=48, other_line_len=44, min_grouping_size=3):
+def vlanlist_to_config(
+    vlan_list: t.List[int],
+    first_line_len: t.Optional[int] = 48,
+    other_line_len: t.Optional[int] = 44,
+    min_grouping_size: t.Optional[int] = 3,
```
Kircheneer, line 13: All of these probably shouldn't be optional, then we can also get rid of some of the ignores below

jeffkala:
@Kircheneer thanks so much for the thorough review, the detail is incredibly helpful.  I'll work through getting everything updated and retested.

jeffkala:
This will be closed, all these changes are also in #125

itdependsnetworks:
Closing in favor of #125

PR #125, opened 2022-06-29
 Add type hints to the whole project and mypy testing setup and CI

Kircheneer:
Addressing review comments from #123


.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
```
itdependsnetworks, line 1: Can you please remove this, and also add .idea to the gitignore?

netutils/config/compliance.py
```diff
@@ -367,19 +380,19 @@ def find_unordered_cfg_lines(intended_cfg, actual_cfg):
...
-        feature (dict): Feature name and cfg lines that should be parsed.
-        device_cfg (str): Device configuration.
-        network_os (str): Device network operating system that is in parser_map keys.
+        feature: Feature name and cfg lines that should be parsed.
+        device_cfg: Device configuration.
+        network_os : Device network operating system that is in parser_map keys.
 
     Returns:
-        list: The hash report data mapping file hashes to report data.
+        he hash report data mapping file hashes to report data.
```
itdependsnetworks, line 395: One letter too far

netutils/config/compliance.py
```diff
@@ -328,15 +341,15 @@ def feature_compliance(feature, backup_cfg, intended_cfg, network_os):
     else:
         raise  # pylint: disable=misplaced-bare-raise
 
-    return feature_data
+    return feature_data  # type: ignore
```
itdependsnetworks, line 344: This still need to be ignored?

netutils/config/parser.py
```diff
@@ -49,49 +60,54 @@ def __init__(self, config):
         super(BaseSpaceConfigParser, self).__init__(config)
 
     @property
-    def indent_level(self):
+    def banner_end(self) -> str:
```
itdependsnetworks, line 63: How is this diff then what it was inherited from?

netutils/config/parser.py
```diff
@@ -328,7 +345,12 @@ def __init__(self, config):
         super(BaseBraceConfigParser, self).__init__(config)
 
     @property
-    def config_lines_only(self):
+    def banner_end(self) -> str:
```
itdependsnetworks, line 348: Same as previous

netutils/interface.py
```diff
@@ -119,15 +122,21 @@ def canonical_interface_name(interface, addl_name_map=None, verify=False):
         name_map.update(addl_name_map)
     # check in dict for mapping
     if name_map.get(interface_type):
-        long_int = name_map.get(interface_type)
+        long_int = name_map.get(interface_type, "")
```
itdependsnetworks, line 125: This should just be a `name_map[interface_type]` since it is checked if it exists already

netutils/ip.py
```diff
@@ -225,16 +227,16 @@ def netmask_to_cidr(netmask):
...
 
     **IPv4** only.  For IPv6, please use `cidr_to_netmaskv6`.
 
     Args:
-        cidr (int): A CIDR value.
+        cidr: A CIDR value.
 
     Returns:
-      netmask (str): Decimal format representation of CIDR value.
+        Decimal format representation of CIDR value.
```
itdependsnetworks, line 239: I see the point, but I think we should change the code to assign a variable as netmask and return it. Additionally spacing seems off

netutils/config/parser.py
Kircheneer: Sorry I don't understand - do you mean why the signature of this function is different than what it is inheriting from? If so, I don't know which exact function you're referring to because to me it seems like the signatures match here and also Mypy [should catch this](https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides)

netutils/ip.py
Kircheneer: 1. There was mix of two-space and four-space for `Returns` in the docstrings, I unified this which led to this diff
2. The conversion to a netmask would be a breaking change which I would prefer not to have in this MR. Do you want me to create an issue for this?

netutils/config/compliance.py
Kircheneer: Fixed!

netutils/interface.py
Kircheneer: Fixed!

netutils/password.py
```diff
@@ -167,16 +172,16 @@ def decrypt_type7(encrypted_password):
...
-        unencrypted_password (str): A password that has not been encrypted, and will be compared against.
-        salt (str, optional): A random set of characters that can be set by the operator. Defaults to random generated one.
-        salt_len (int, optional): The number of random set of characters, when not manually set. Defaults to 4.
+        unencrypted_password: A password that has not been encrypted, and will be compared against.
+        salt: A random set of characters that can be set by the operator. Defaults to random generated one.
+        salt_len: The number of random set of characters, when not manually set. Defaults to 4.
 
     Returns:
-        string: The encrypted password.
+        str: The encrypted password.
```
itdependsnetworks, line 184: Why keep this one as `str`?

netutils/config/parser.py
Kircheneer: Sorry I see now what you mean, I removed these.

netutils/ip.py
Kircheneer: Removed the `netmask:` bit from both so its consistent.

netutils/password.py
Kircheneer: I missed it, thank you.

Kircheneer:
> Interesting, failing only in 3.7 with:
> 
> ```
> ************* Module netutils.config.parser
> netutils/config/parser.py:47:0: W0223: Method 'banner_end' is abstract in class 'BaseConfigParser' but is not overridden (abstract-method)
> netutils/config/parser.py:329:0: W0223: Method 'banner_end' is abstract in class 'BaseConfigParser' but is not overridden (abstract-method)
> ```
> 
> This may be why you added banner_end, but I don't see why it is needed.

Probably, I just ran `mypy` before I submitted that commit.. I gave it a quick google and found [this](https://github.com/PyCQA/pylint/issues/3098) - this seems to suggest that adding a metaclass is the way to go. So I think we can either
- subclass all the classes which are de-facto abstract from `abc.ABC` or
- add empty (i.e. useless) implementations for these methods as I had before.

I would say the latter is better because we save on an import and a little complexity. What do you think @itdependsnetworks ?

itdependsnetworks:
I would just disable abstract-method / W0223 on those classes.

netutils/config/clean.py
```diff
@@ -52,12 +52,12 @@ def clean_config(config, filters):
     return config
 
 
-def sanitize_config(config, filters):
+def sanitize_config(config: str, filters: t.Optional[t.List[t.Dict[str, str]]] = None) -> str:
```
itdependsnetworks, line 55: Why set it to none?

Kircheneer: As per docstring this needs to default to empty and due to the mutability of default arguments we cannot default to an empty list, therefore None and this:

```
if not filters:
    filters = []
```

itdependsnetworks: I would change the doc string to meet the code, not the other way around

Kircheneer: Sorry I missed this one, because of this: https://docs.python-guide.org/writing/gotchas/

netutils/interface.py
```diff
@@ -118,16 +121,22 @@ def canonical_interface_name(interface, addl_name_map=None, verify=False):
     if isinstance(addl_name_map, dict):
         name_map.update(addl_name_map)
     # check in dict for mapping
-    if name_map.get(interface_type):
-        long_int = name_map.get(interface_type)
+    if interface_type in name_map:
```
itdependsnetworks, line 124: can we do `if interface_type in name_map.keys():`? I think the explicit speaks to the future developer looking at this seeing that it is a dictionary.

netutils/time.py
```diff
@@ -64,6 +66,7 @@ def uptime_string_to_seconds(uptime_string):
 
     uptime_seconds = 0
     for time_interval, value in TIME_MAPPINGS:
-        if uptime_dict.get(time_interval):
-            uptime_seconds += int(uptime_dict.get(time_interval)) * value
+        time_interval_as_int = uptime_dict.get(time_interval)
```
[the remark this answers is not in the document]
Kircheneer: Mypy does not keep state on variables for a lot of patterns. As a consequence, it is not able to infer that `type(uptime_dict.get(time_interval)) == int != Optional[int]` after that if clause. The way it's written here, mypy knows that `type(time_interval_as_int) == Optional[int]` and then we perform explicit optional (i.e. `None`) handling after that.

netutils/vlan.py
```diff
@@ -28,14 +33,16 @@ def vlanlist_to_config(vlan_list, first_line_len=48, other_line_len=44, min_grou
...
 
-    def build_final_vlan_cfg(vlan_cfg):
+    def build_final_vlan_cfg(vlan_cfg: str) -> t.List[str]:
         if len(vlan_cfg) <= first_line_len:
             return [vlan_cfg]
 
         # Split VLAN config if lines are too long
         first_line = re.match(f"^.{{0,{first_line_len}}}(?=,)", vlan_cfg)
+        if not first_line:
+            raise ValueError(f"Parsing VLAN list with first_line_len={first_line_len} yielded no regex results.")
```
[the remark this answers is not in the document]
Kircheneer: @jeffkala I'm going to need your help on this one because you understand the business logic better, can you give a suggestion to make this error message a little friendlier to everyone? I think not even the software engineer will be able to do something with this 🤦🏻

netutils/interface.py
Kircheneer: Hm I personally don't like it - shouldn't we be able to mentally infer for the `_map` suffix that this variable is a dictionary? Also explicitly calling `.keys()` for `in` checks is not a pattern I have seen before as far as I can remember.

netutils/time.py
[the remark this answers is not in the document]
itdependsnetworks: thanks.

netutils/interface.py
Kircheneer: Closing after slack conversation.

itdependsnetworks:
Addressed in #129

Issue #124, opened 2022-06-26
Config Parsing Helper


itdependsnetworks:
### Environment
* netutils version:  1.1

### Proposed Functionality

Provided convinence methods to parse through a standard Ios-like (and potentially others) configurations. The idea being similar to CiscoConfParse in scope, but ideally to provide a few key features. 

The ability to navigate nested configuration for a known path:
```
root_obj.get_path(["router bgp 45000", "address-family ipv4 unicast"])
# Returns 
  neighbor 192.168.1.2 activate
  network 172.17.1.0 mas
```

The ability to get configurations with parent context:
```
root_obj.get_path_with_parents(["router bgp 45000", "address-family ipv4 unicast"])
# Returns 
router bgp 45000
 address-family ipv4 unicast
  neighbor 192.168.1.2 activate
  network 172.17.1.0 mas
```

The ability to get configurations that match:
```
root_obj.get_multpile_parents(["interface GigabitEthernet.*"])
# Returns 
[ """interface GigabitEthernet0/1/0
 mac-address 00e0.1476.7600
 no ip address
!""", 
"""interface GigabitEthernet0/1/1
 mac-address 00e0.1476.7601
 no ip address
!"""]
```

The ability to get configurations that contains:
```
root_obj.get_contained(["interface GigabitEthernet.*"], "channel-group 1")
# Returns 
[ """interface GigabitEthernet0/1/0
 mac-address 00e0.1476.7600
 no ip address
 channel-group 1
!""", 
"""interface GigabitEthernet0/1/1
 mac-address 00e0.1476.7601
 no ip address
 channel-group 1
!"""]
```

And similar

### Use Case

As an operator, I want the ability to find configurations based on various different characteristics. Such that I can quickly make logical choices, populate a SoT, find otherwise difficult to find configurations (offline), and quickly interrogate configurations across a large amount of data, just to name a few.

jeffkala:
Would the idea be to rework how the parsers work, or for these to be its own logic?

itdependsnetworks:
I think it should just be adding convenience methods.

mundruid:
Functions to be added:
   def find_blocks():
    def find_children():
    def find_children_w_parents():
    def find_interface_objects():
    def find_lineage():
    def find_lines():
    def find_object_branches():
    def find_objects():
    def find_objects_dna():
    def find_objects_w_all_children():
    def find_objects_w_child():
    def find_objects_w_missing_children():
    def find_objects_w_parents():
[...]

PR #140, opened 2022-08-03
Adds utility functions ios

mundruid:
Adds:
- `get_path`
- `get_path_with_children`

Addresses partially issue: https://github.com/networktocode/netutils/issues/124

### DIFF
--- a/netutils/config/parser.py
+++ b/netutils/config/parser.py
@@ -3,6 +3,7 @@
 
 import re
 import typing as t
+import itertools as it
 from collections import namedtuple
 
 from netutils.banner import normalise_delimiter_caret_c
@@ -587,6 +588,73 @@ class IOSConfigParser(CiscoConfigParser, BaseSpaceConfigParser):
         super(IOSConfigParser, self).build_config_relationship()
         self._update_same_line_children_configs()
         return self.config_lines
+
+    def _get_groups(self, pattern: str) -> t.Any:
+        """Groups children based on parent pattern."""
+        children_list = []
+        for line in self.config_lines:
+            for parent in line.parents:
+                if re.match(pattern, parent):
+                    children_list.append(line)
+        grouped_data = it.groupby(children_list, key=lambda x: x.parents)  # type: ignore
+        return grouped_data
+
+    def get_path(self, pattern: str) -> t.List[str]:
+        """Returns configuration part for a specific pattern not including parents.
+
+        Args:
+            pattern: pattern that describes for parent.
+
+        Returns:
+            configuration under that parent pattern.
+
+        Example:
+            >>> config = '''
+            ... router bgp 45000
+            ...   address-family ipv4 unicast
+            ...    neighbor 192.168.1.2 activate
+            ...    network 172.17.1.0 mas'''
+            >>> bgp_conf = IOSConfigParser(str(config)).get_path("router bgp")
+            >>> print(bgp_conf)
+            ['  address-family ipv4 unicast', '   neighbor 192.168.1.2 activate', '   network 172.17.1.0 mas']
+        """
+        children = []
+        grouped_data = self._get_groups(pattern)
+        for _, grp in grouped_data:
+            for line in list(grp):
+                children.append(line.config_line)
+        return children
+
+    def get_path_with_parents(self, pattern: str) -> t.List[str]:
+        """Returns configuration part for a specific pattern including parents and children.
+
+        Args:
+            pattern: pattern that describes for parent.
+
+        Returns:
+            configuration under that parent pattern.
+
+        Example:
+            >>> config = '''
+            ... router bgp 45000
+            ...   address-family ipv4 unicast
+            ...    neighbor 192.168.1.2 activate
+            ...    network 172.17.1.0 mas'''
+            >>> bgp_conf = IOSConfigParser(str(config)).get_path_with_parents("router bgp")
+            >>> print(bgp_conf)
+            ['router bgp 45000', '  address-family ipv4 unicast', '   neighbor 192.168.1.2 activate', '   network 172.17.1.0 mas']
+        """
+        parents_and_children = []
+        previous_parent = ""
+        grouped_data = self._get_groups(pattern)
+        for key, grp in grouped_data:
+            # special case for nested parents, we do not want to repeat these multiple times
+            if previous_parent != key[0]:
+                parents_and_children.append(key[0])
+            previous_parent = key[0]
+            for line in list(grp):
+                parents_and_children.append(line.config_line)
+        return parents_and_children
 
 
 class NXOSConfigParser(CiscoConfigParser, BaseSpaceConfigParser):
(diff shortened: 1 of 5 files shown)


qduk (approved):
Thanks for the docs fix!

netutils/config/parser.py
```diff
@@ -331,6 +331,100 @@ def build_config_relationship(self) -> t.List[ConfigLine]:
...
+
+        Returns:
+            configuration under that parent pattern.
+        Example:
+            >>> config = '''
+            ... router bgp 45000
+            ...   address-family ipv4 unicast
+            ...    neighbor 192.168.1.2 activate
+            ...    network 172.17.1.0 mas'''
+            >>> bgp_conf = BaseSpaceConfigParser(str(config)).find_all_children(pattern="router bgp", match_type="regex")
```
jeffkala, line 362: I feel like an example with `match_type="regex"` the pattern should be a regex of some sort to properly demonstrate.

netutils/config/parser.py
```diff
@@ -331,6 +331,100 @@ def build_config_relationship(self) -> t.List[ConfigLine]:
...
+            match_type (optional): Exact or regex. Defaults to "exact".
+
+        Returns:
+            configuration under that parent pattern.
+        Example:
+            >>> config = '''
+            ... router bgp 45000
+            ...   address-family ipv4 unicast
+            ...    neighbor 192.168.1.2 activate
+            ...    network 172.17.1.0 mas'''
```
jeffkala, line 361: Not following this command? Missing `mask`?

netutils/config/parser.py
```diff
@@ -331,6 +331,100 @@ def build_config_relationship(self) -> t.List[ConfigLine]:
...
```
jeffkala, line 394: same as above

netutils/config/parser.py
```diff
@@ -331,6 +331,100 @@ def build_config_relationship(self) -> t.List[ConfigLine]:
...
+            >>> bgp_conf = BaseSpaceConfigParser(str(config)).find_children_w_parents(parent_pattern="router bgp", child_pattern="  address-family", match_type="regex")
+            >>> print(bgp_conf)
+            ['  address-family ipv4 unicast', '   neighbor 192.168.1.2 activate', '   network 172.17.1.0 mas']
```
jeffkala, line 397: Shouldn't `router bgp 45000` be returned since its the parent?  Also similar to above perhaps I'm not understanding how the pattern pattern not being a regex is working with `match_type="regex"`.

netutils/config/parser.py
```diff
@@ -331,6 +331,100 @@ def build_config_relationship(self) -> t.List[ConfigLine]:
...
+    # def find_object_branches():
+    # def find_objects():
+    # def find_objects_dna():
+    # def find_objects_w_all_children():
+    # def find_objects_w_child():
+    # def find_objects_w_missing_children():
+    # def find_objects_w_parents():
+    # def find_objects_wo_child():
+    # def find_parents_w_child():
+    # def find_parents_wo_child():
```
jeffkala, line 426: Are these future WIP for this PR?  If not I'd prefer to document these in a issue versus inside the source code.

jeffkala (changes requested):

tests/unit/test_parser.py
```diff
@@ -14,6 +15,16 @@
         parameters.append([_file, network_os])
 
 
+# TODO: add more tests with different patterns
```
itdependsnetworks, line 18: Instead of manually generating these, see the pattern I provided in #141

tests/unit/test_parser.py
```diff
@@ -37,3 +48,27 @@ def test_incorrect_banner_ios():
     )
     with pytest.raises(ValueError):
         compliance.parser_map["cisco_ios"](banner_cfg).config_lines  # pylint: disable=expression-not-assigned
+
+
+@pytest.mark.parametrize("_file, pattern, expected", get_path_parameters)
```
itdependsnetworks, line 53: ok, I see you added it here, but didn't define `get_path_parameters`, in #141 you should see how that is auto populated based on folder structure. The premise is you just put files in specfic locations, and it will dynamically find it

netutils/config/parser.py
mundruid: Changed to `startswith`

netutils/config/parser.py
mundruid: Changed.

netutils/config/parser.py
mundruid: Changed.

netutils/config/parser.py
mundruid: Changed to `startswith`. I think in this case we should not return the parent

netutils/config/parser.py
mundruid: Yes, future WIP. Will document elsewhere. Deleted.

tests/unit/test_parser.py
mundruid: I am not sure what you mean here.

tests/unit/test_parser.py
mundruid: I still do not understand what you are asking. This is a set of parameters for the unit test. Did you want to dynamically populate it for different platforms? You can do this with parametrize.

itdependsnetworks: I took this from 141, this uses glob to dynamically find files, which populates the variables for paramatized variable. With these two concepts you can create a framework, in which you simply drop files in a certain location, and they are automatically tests.

```
base_parameters = []
find_all_children_parameters = []
find_children_w_parents_parameters = []
for network_os in list(compliance.parser_map.keys()):
    for _file in glob.glob(f"{MOCK_DIR}/base/{network_os}/*{TXT_FILE}"):
        base_parameters.append([_file, network_os])
    for _file in glob.glob(f"{MOCK_DIR}/find_all_children/{network_os}/*{TXT_FILE}"):
        find_all_children_parameters.append([_file, network_os])
    for _file in glob.glob(f"{MOCK_DIR}/find_children_w_parents/{network_os}/*{TXT_FILE}"):
        find_children_w_parents_parameters.append([_file, network_os])
```

mundruid: Addressed.

tests/unit/test_parser.py
mundruid: Addressed see below.

itdependsnetworks:
I don't understand how there is actually tests here? There just seems to be configurations, but nothing saying that configuration 1 should be parsed as X.

mundruid:
> I don't understand how there is actually tests here? There just seems to be configurations, but nothing saying that configuration 1 should be parsed as X.

Are you referring to the unit tests that I wrote? I parse an ios configuration and use the functions that are written: `find_all_children` and `find_children_w_parents`. These two functions are the units under test. Then I compare the output of these functions with the expected output. If the test passes, I consider the unit, i.e. the functions mentioned, tested. I also consider that this is the expected behavior. Are you expecting something different for the unit tests?

itdependsnetworks (approved):

PR #141, opened 2022-08-04
Add example parsers

itdependsnetworks:


itdependsnetworks:
@mundruid take a look at this, spent a little time to setup a pattern I was trying to mention. 2 primary things:

- The methods should be doable with just a few lines of code iterating over the namedtuple ConfigLine
- The parametrization of the globed files in tests

I think most if not all of the parsers from CiscoConfParse can be replicated. This way we do not have to worry about GPL moving forward.

PR #149, opened 2022-09-15
Add Napalm Getter feature

### LINKED ISSUE #95: Napalm Supported Getter Matrix (as code)
### Environment
* netutils version:  future

### Proposed Functionality
A mapper function to be able to find supported napalm getters in a programatic way.  Would essentially map https://napalm.readthedocs.io/en/latest/support/#getters-support-matrix into a dictionary of sorts.

### Use Case
UI interface of an application allows you to utilize Napalm as a connection type, it then has a drop down of supported getters based on the platform selected.

jeffkala:
fix #95

### DIFF
--- a/netutils/utils.py
+++ b/netutils/utils.py
@@ -1,6 +1,18 @@
 """Utilities for the netutils library."""
+
+import inspect
 import typing as t
 from importlib import import_module
+
+from netutils.lib_mapper import NAPALM_LIB_MAPPER
+
+try:
+    from napalm import get_network_driver
+    from napalm.base.exceptions import ModuleImportError
+except ImportError:
+    HAS_NAPALM = False
+else:
+    HAS_NAPALM = True
 
 _JINJA2_FUNCTION_MAPPINGS = {
     "asn_to_int": "asn.asn_to_int",
@@ -88,3 +100,37 @@ def jinja2_convenience_function() -> t.Dict[str, t.Callable[..., t.Any]]:
         function_object = getattr(imported_module, function_name)
         result[jinja2_function_name] = function_object
     return result
+
+
+def get_napalm_getters() -> t.Dict[str, t.Dict[str, bool]]:
+    """Utility to return a dictionary of napalm getters based on install napalm version.
+
+    Returns:
+        Keys are OS and values are a dictionary of supported napalm getters.
+
+    Raises:
+        ImportError: If optional dependency Napalm is not installed.
+
+    Example:
+        >>> from netutils.utils import get_napalm_getters
+        >>> napalm_getters = get_napalm_getters()
+        >>> napalm_getters["eos"]["get_arp_table"]
+        >>> True
+        >>> napalm_getters["eos"]["get_ipv6_neighbors_table"]
+        >>> False
+    """
+    if HAS_NAPALM:
+        napalm_dict = {}
+        oses = NAPALM_LIB_MAPPER.keys()
+        for my_os in oses:
+            try:
+                get_network_driver(my_os)
+            except ModuleImportError:
+                continue
+            napalm_dict[my_os] = {}
+            for getter in inspect.getmembers(get_network_driver(my_os), predicate=inspect.isfunction):
+                if getter[0].startswith("get_"):
+                    state = False if getter[1].__module__ == "napalm.base.base" else True
+                    napalm_dict[my_os][getter[0]] = state
+        return napalm_dict
+    raise ImportError("Napalm must be install for this function to operate.")


itdependsnetworks:
Can you add the optional dependencies, our json schema repo should show how that is done. Need to add tests as well

netutils/utils.py
```diff
@@ -88,3 +100,37 @@ def jinja2_convenience_function() -> t.Dict[str, t.Callable[..., t.Any]]:
...
+
+    Example:
+        >>> from netutils.utils import get_napalm_getters
+        >>> napalm_getters = get_napalm_getters()
+        >>> napalm_getters["eos"]["get_arp_table"]
+        >>> True
+        >>> napalm_getters["eos"]["get_ipv6_neighbors_table"]
+        >>> False
+    """
+    if HAS_NAPALM:
```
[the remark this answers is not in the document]
jeffkala: changed this to have the imports inside the function. that way they're only loaded if the function is called.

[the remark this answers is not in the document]
itdependsnetworks: I think the other pattern is more common

netutils/utils.py
```diff
@@ -1,7 +1,11 @@
 """Utilities for the netutils library."""
```
itdependsnetworks, line 1: I don't think this is the right file for this. We should have something like parsers or similar.

netutils/utils.py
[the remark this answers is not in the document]
jeffkala: More common, but this at least will only load when that one function is called, also the nice trade off of doing it within the function is I can mock these imports in the tests without needing to have `napalm` as an testing dependency in the pyproject.toml.

tests/unit/test_utils.py
```diff
@@ -25,7 +28,7 @@
 
 _EXCLUDED_DECORATOR_FUNCTIONS = ["wraps", "total_ordering", "abstractmethod"]
 
-_EXCLUDED_FUNCTIONS = ["jinja2_convenience_function", "import_module"]
+_EXCLUDED_FUNCTIONS = ["jinja2_convenience_function", "import_module", "get_network_driver"]
```
itdependsnetworks, line 31: Why is this function excluded?

netutils/utils.py
jeffkala: I almost think something like `lib_helpers` or something would be good. Thoughts?

netutils/utils.py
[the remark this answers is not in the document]
itdependsnetworks: Python does not load in a greedy fashion, this module will only load if you call the module anyway, take a look a this:

```
>>> import netutils
>>> dir(netutils)
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '__version__']
>>>
```
If this was greedy, or we explicitly set to set in the init file, we would see things methods like `asn`, `vlans`, etc..


Once this change was introduced https://github.com/networktocode/netutils/commit/cb249f088bbfbdec2addf7c6634c8caab7e0e8a2#diff-14c4cb1fd1f636a25adaea989f0875f63e36e3b18ae2d9f83bd7036975fe9509 this would not load automatically 

I am ultimately not sure that not installing the dependencies in the test environment would pass the POLA. That being said, I've not dealt with this situation too much, so could be good to reach out internally to gather ideas.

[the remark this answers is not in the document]
jeffkala: Its easier for testing etc to have napalm as a development dependency. Makes testing much easier as well, now that i've read about POLA I agree that i'd expect this to be part of the test environment dep.

tests/unit/test_utils.py
jeffkala: Its a method imported from napalm, we just use it. I don't see what it would offer as a filter.

itdependsnetworks: doh`

pyproject.toml
```diff
@@ -39,12 +39,13 @@ pytest = "*"
 pyyaml = "*"
 pydocstyle = "*"
 m2r2 = "*"
-sphinx = "*"
 sphinx-rtd-theme = "*"
 toml = "*"
 yamllint = "*"
 mypy = "^0.961"
 black = "*"
+napalm = "^4.0.0"
```
itdependsnetworks, line 47: I think this should also be in `[tool.poetry.extras]`. dev-dependencies are for your testing strategy, but extras allows you to do something like `pip install netutils[napalm]`

jeffkala:
```
▶ poetry add -E netlibs napalm
The following packages are already present in the pyproject.toml and will be skipped:

  • napalm

If you want to update it to the latest compatible version, you can use `poetry update package`.
If you prefer to upgrade it to the latest available version, you can use `poetry add package@latest`.

Nothing to add.
```

jeffkala: @itdependsnetworks better way to do this?

itdependsnetworks: Do you mind pinging ask-python internally? I haven't set this up personally before. Phillip may have done it on jsonschema as an example.

tasks.py
```diff
@@ -46,40 +49,71 @@ def is_truthy(arg):
...
+    """Gets the name of the container image to use.
+
+    Args:
+        with_napalm (bool): Get name of container image with Napalm installed.
+
+    Returns:
+        str: Name of container image. Includes tag.
+    """
+    if with_napalm and not os.getenv("GITHUB_ACTION", None):
+        name = f"{IMAGE_NAME}:{IMAGE_VER}-{NAPALM_PACKAGE}{NAPALM_VER}"
```
itdependsnetworks, line 62: Will this scale if we have one for ntc-templates, pyats, etc?

jeffkala: looking into poetry's new `groupings` might allow us to do this easier since ultimately this is more of development related issues.

itdependsnetworks: Yea, good point. I think we can have a "full" version, and another no-dependency version. This is just for development environment afterall.

itdependsnetworks:
I think that we need to document how to install the optional dependencies and when to install them

jeffkala:
Will work on getting docs updated.

jeffkala:
should be ready now.

README.md
```diff
@@ -43,12 +44,23 @@ Option 1: Install from PyPI.
 $ pip install netutils
 ```
 
-Option 2: Install from a GitHub branch, such as develop as shown below.
+Option 2: Install from PyPI with optional requirements.
+
+```bash
+$ pip install netutils[napalm]
```
itdependsnetworks, line 50: Something like 

```
'''
$ pip install netutils
'''

Install with optional dependencies.

'''
$ pip install netutils[napalm]
'''
```

Please note I used single quotes instead of back ticks in example, but would expect that you would use back ticks in real implementation.

README.md
```diff
@@ -223,6 +235,7 @@ Each command can be executed with `invoke <command>`. Each command also has its
   pylint             Run pylint code analysis.
   pydocstyle         Run pydocstyle to validate docstring formatting adheres to NTC defined standards.
   pytest             Run pytest for the specified name and Python version.
+  pytest-without-optionals   This will run pytest only to assert the correct errors are raised when optional dependencies are not installed.
```
itdependsnetworks, line 238: Can we make it, pytest and pytest-all? Would need to flip the logic though of course.

jeffkala: meaning `pytest` is without options, all is with them?

itdependsnetworks: yes.

.github/workflows/ci.yml
```diff
@@ -157,6 +172,44 @@ jobs:
         run: "poetry run invoke pylint"
     needs:
       - "build"
+  pytest_all:
```
itdependsnetworks, line 175: I don't think we need to separate out, the ci can just run with pytest-all, no? Otherwise, creating more steps in the github actions runs

docs/admin/install.md
```diff
@@ -6,8 +6,19 @@ Option 1: Install from PyPI.
...
+
+Option 3: Install from a GitHub branch, such as develop as shown below.
 
 ```bash
 $ pip install git+https://github.com/networktocode/netutils.git@develop
-```
\ No newline at end of file
+```
+
+### Optional Dependencies
```
itdependsnetworks, line 21: Can you add something similar to faq?

.github/workflows/ci.yml
jeffkala: right now `pytest` only runs 1 test, and that is the check to make sure the exception is raised on the get_napalm_getters() when napalm is not installed. `pytest-all` runs every other test including testing the actual get_napalm_getters() functionality when napalm is installed. I must be missing something on what you're asking.

itdependsnetworks: something _like_
```
    strategy:
      fail-fast: true
      matrix:
        python-version: ["3.7", "3.8", "3.9", "3.10"]
        testing-type: "pytest_all"
        include:
          - python-version: "3.10"
            testing-type: "pytest"
```

pyproject.toml
```diff
@@ -38,7 +42,6 @@ pytest = "*"
 pyyaml = "*"
 pydocstyle = "*"
 m2r2 = "*"
-sphinx = "*"
 sphinx-rtd-theme = "*"
```
itdependsnetworks, line 45: can likely remove as well, (missed by me)

itdependsnetworks: along with m2r2

tests/unit/test_lib_helpers_optionals.py
```diff
@@ -0,0 +1,59 @@
+"""Test for the lib_helpers definitions when optional packages are installed."""
+from unittest import mock
+
+import pytest
+from netutils.lib_helpers import get_napalm_getters
+
+
+def test_get_napalm_getters_napalm_installed_default():
+    pytest.importorskip("napalm")
```
itdependsnetworks, line 9: nice find.

.github/workflows/ci.yml
```diff
@@ -246,6 +192,10 @@ jobs:
         run: "docker image ls"
       - name: "Run Tests"
         run: "poetry run invoke pytest"
+      - name: "Install Napalm"
+        run: "pip install napalm"
+      - name: "Run Optional Tests"
+        run: "find tests/ -name 'test_lib_helpers_optionals.py' | xargs pytest -vv"
```
itdependsnetworks, line 198: `pytest test_lib_helpers_optionals.py` or similar should work.

pyproject.toml
jeffkala: removing m2r2 is good, once I remove sphinx-rtd-theme I can no longer build.

jeffkala: see initial pipeline failure

jeffkala: removed m2r2, left sphinx-rtd-theme for now until we determine what is using this dependency and while builds start to fail once its removed.

itdependsnetworks (approved):

jeffkala:
@qduk want to review?

qduk (approved):
LGTM!

Issue #88, opened 2021-12-20
Readme for adding a parser.


jeffkala:
### Environment
* netutils version:  

A readme file in configs that provides the proper understanding of how to write a new parser would be helpful.

### Use Case
External contributor wants to add support for a new parser, it'd be nice for a getting started how to guide.

PR #151, opened 2022-10-08
Add jinja2_convenience_function docs and how to add a config parser

### LINKED ISSUE #150: Document the out of the box jinja2_convenience_function syntax
### Environment
* netutils version:  1.2.0

### Proposed Functionality

Document the jinja2_convenience_function filter functions similar to the way each function is documented.

### Use Case

This would allow users to better understand how to leverage jinja2_convenience_function filters in their jinja.

itdependsnetworks:
Fixes #150 and #88

### DIFF
--- a/docs/source/netutils/configs/index.rst
+++ b/docs/source/netutils/configs/index.rst
@@ -12,14 +12,20 @@ Configs
     :members:
 
 Edge Cases
-==============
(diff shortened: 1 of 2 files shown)


itdependsnetworks:
Replaced by #152

Issue #86, opened 2021-12-09
Future Parser Support List


jeffkala:
List of Parser that should be created.
- [ ] IOS-XR
- [ ] Aruba
- [ ] Checkpoint

jdrew82:
Also need to add following:

jdrew82:
A few more to add:
- [ ] A10
- [ ] NetScaler
- [ ] Netscreen

angely-dev:
Any updates on Huawei parser?

Huawei config is a `BaseSpaceConfigParser` with:

* `comment_chars` as `#`
* `banner_start` as `header login information`
* A banner starts and ends with `"`

A one-line banner:

```
[device]header login information "My one-line banner"
```

A multi-line banner:

```
[device]header login information "
The banner text supports 480 characters max, including the start and the end character.If you want to enter more than this, use banner file instead.Input banner text, and quit with the character '"':
==============================
My multi-line banner
==============================
"
```

A sample config file:

```
#
interface GigabitEthernet0/0/4
 undo portswitch
 undo shutdown
 l2 binding vsi CUST
#
interface GigabitEthernet0/0/5
 undo portswitch
 undo shutdown
 l2 binding vsi CUST
#
bgp 34177
 router-id 1.1.1.1
 ipv4-family vpn-instance ADMIN
  import-route static
#
header login information "==============================
My multi-line banner
==============================
"
#
```

Is this enough for you to implement it, or do you prefer a PR?

Thanks!

itdependsnetworks:
PR would be best, can see some hints here: https://netutils.readthedocs.io/en/latest/dev/dev_config/#new-parsers

angely-dev:
Thanks @itdependsnetworks.

So using this simple snippet:

```py
# netutils/config/parser.py

class HuaweiVRPConfigParser(BaseSpaceConfigParser):
    """Huawei VRP implementation of ConfigParser Class."""

    comment_chars: t.List[str] = ["#"]
    banner_start = ["header login information"]
    banner_end = '"'
```

I got this diff output:

```
interface GigabitEthernet0/0/2
 l2 binding vsi CUST
header login information "==================================================
my multi-line banner
==================================================^C
```

The banner end delimiter `"` gets replaced by `^C`. It is done by the [generic space config parser #L215](https://github.com/networktocode/netutils/blob/develop/netutils/config/parser.py#L215). Is this normal behavior? I can override the `_build_banner` method but I doubt this is the guideline.

[...]

PR #153, opened 2022-10-11
Citrix NetScaler Parser

jdrew82:
This PR is for adding a Citrix NetScaler parser. I've included an example for the unit test and confirmed tests are passing. Please let me know if there are any other tests I need to write for this. This should address one of the requests in #86.


itdependsnetworks:
Do you mind rebasing and running tests? You will have to run the development script in the root of the directory to regen one of the docs.

tests/unit/mock/config/compliance/compliance/citrix_netscaler/netscaler_basic_feature.py
```diff
@@ -0,0 +1,3 @@
+features = [
+    {"name": "user", "ordered": False, "section": ["set system user "]},
```
itdependsnetworks, line 2: Can you add a test for `add system cmdPolicy`? This has regex in the line and would be good to make sure nothing gets tripped up.

itdependsnetworks: Also one with an add and bind, such as `add ssl` and `bind ssl`

tests/unit/mock/config/compliance/compliance/citrix_netscaler/netscaler_basic_backup.txt
```diff
@@ -0,0 +1,85 @@
...
+set snmp alarm IP-CONFLICT -severity Warning
+set snmp alarm MEMORY -thresholdValue 95 -normalValue 35 -severity Critical
+set snmp alarm POWER-SUPPLY-FAILURE -severity Minor
+set snmp alarm SSL-CARD-FAILED -severity Minor
+set snmp alarm SSL-CERT-EXPIRY -severity Warning
+add snmp view READ 1 -type included
+add snmp group NETMON-GROUP authpriv -readViewName READ
+add snmp user monitoring -group NETMON-GROUP -authType SHA -authpasswd abcdef1234 -privType AES -privpasswd abcdef1234
+add ssl cipher XX-CIPHER-GROUP_1.0_v01
+bind ssl cipher XX-CIPHER-GROUP_1.0_v01 -cipherName TLS1.2-ECDHE-RSA-AES256-GCM-SHA384
```
itdependsnetworks, line 50: @jeffkala mentioned, and just want to confirm, would we ever want a parent and child relationship for add & bind. I don't think so, but just want to think of it.

jeffkala:
Also would be worthwhile based on the result of @itdependsnetworks comment above, if we don't want parent/child we should put a note similar to this for NetScaler.
https://netutils.readthedocs.io/en/latest/dev/dev_config/#fortinet-fortios-parser

jdrew82:
@jeffkala I've made a note in the documentation about the lack of parent/child relationships due to the configuration structure. I've also gone ahead and added the tests requested @itdependsnetworks.

tests/unit/mock/config/compliance/compliance/citrix_netscaler/netscaler_basic_feature.py
jdrew82: Done

docs/dev/dev_config.md
```diff
@@ -50,22 +54,21 @@ Documented use cases that are actual configuration on a network device are consi
 
 ## New Parsers
 
-
 There are a series of considerations documented below, when developing a new parser.
 
 - Creation of a new class that must be created in `netutils/config/parser.py` file.
 - Creation of a parser class that inherits from the class `BaseConfigParser` in the Python Method Resolution Order (MRO).
-    - In nearly all cases should inherit directory off of `BaseSpaceConfigParser` or `BaseBraceConfigParser`.
```
itdependsnetworks, line 58: I don't think this is rendered correctly in mkdocs, I think this needs to be reverted.

jdrew82: I'm assuming you're meaning that this particular line shouldn't be indented?

tests/unit/mock/config/compliance/compliance/citrix_netscaler/netscaler_basic_backup.txt
jeffkala: @jdrew82 think I was thinking specifically like this cipher group `XX-CIPHER-GROUP_1.0_v01` the add accomplishes only a subset of the configuration.  Just wondering if the `add` line should be the parent and the `bind` lines would be children since they appear to be related.

jdrew82: I think it might be difficult to make add/bind as a global parent/child relationship when it comes to NetScaler. If I remember correctly, the `add` line is done to create an object and the `bind` line is to apply it to something else typically. In this specific case, it's creating a cipher group and adding ciphers into it by the looks of it. However, when it comes to something like a VIP you'd be binding the profile to the virtual server.

docs/dev/dev_config.md
itdependsnetworks: I am asking to keep as is, I do not think it will render correctly. Under Github markdown 2 spaces is fine, mkdocs, I don't think it is. I could be wrong, but can you provide a screenshot?

jdrew82: I've reverted the indent and I believe it should be as it was before with the NetScaler note added.

tests/unit/mock/config/compliance/compliance/citrix_netscaler/netscaler_basic_backup.txt
itdependsnetworks: I am good with it.

itdependsnetworks (approved):

jeffkala (approved):

PR #165, opened 2022-11-04
WIP: Fixing flatbot git issue

qduk:


jeffkala:
Still feel like this should only be run on PRs into `main` simply because that's when we typically go do a new release. My mind can be changed.

qduk:
@jeffkala Yeah, I agree it should run on a PR to main. I think this should fix it. This tells the checkout action to checkout the PR head. https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit. This should prevent it from going into detached head state.

PR #171, opened 2022-11-10
Allow proto mapper to be auto-generated

itdependsnetworks:


jeffkala:
bump @itdependsnetworks

itdependsnetworks:
Should be good now, but would like for @jeffkala and @qduk to review the flat bot stuff.

.github/workflows/proto_filepull.yml
```diff
@@ -0,0 +1,50 @@
...
+  schedule:
+    - cron: "0 8 * * 4"
+
+jobs:
+  data_gathering:
+    runs-on: "ubuntu-latest"
+    env:
+      BRANCH_NAME: "PROTO_Updates"
+    steps:
+      # Checkout repo
```
jeffkala, line 14: I'd check on https://github.com/networktocode/netutils/pull/197 to make sure we follow suit here.

flat_postprocess/protocol_postprocess.py
```diff
@@ -0,0 +1,37 @@
...
+import os
+import sys
+from urllib.request import urlopen
+
+
+PROTOCOL_MAPPINGS = {}
+
+if __name__ == "__main__":
+    if len(sys.argv) == 3:
+        URL = "https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.csv"
```
jeffkala, line 12: I get a 500 server error when trying to access this url in straight browser. Just making sure there is no typo here.

flat_postprocess/protocol_postprocess.py
```diff
@@ -0,0 +1,37 @@
...
+        reader = csv.reader(file)
+        for row in reader:
+            name = row[0]
+            port = row[1]
+            protocol = row[2]
+            if not port.isnumeric():
+                continue
+            if name and port and protocol:
+                if not PROTOCOL_MAPPINGS.get(name):
+                    PROTOCOL_MAPPINGS[name] = {"port_number": int(port), "protocols": []}
```
jeffkala, line 28: Wondering if we should try to use a ordered dict here instead, this will avoid unnecessary PR updates if the dictionary is just in a new order with the same data?  See comment on this PR for a similar related issue.
https://github.com/networktocode/netutils/pull/199#discussion_r1095088700

jeffkala (changes requested):

flat_postprocess/protocol_postprocess.py
itdependsnetworks: It's odd, I got that once yesterday as well, but just confirmed it is good.

flat_postprocess/protocol_postprocess.py
itdependsnetworks: The format comes out from iana in numeric order, I think we are ok here.

jeffkala (approved):
Everything looks good to me here.

qduk (approved):

PR #172, opened 2022-11-10
Aruba AOS-CX Parser

lukebates123:
Added Aruba AOS-CX Parser.
I created tests based on the docs here https://netutils.readthedocs.io/en/latest/dev/dev_config/#new-parsers if there are additional tests needed let me know.  This also addresses a request in https://github.com/networktocode/netutils/issues/86


docs/dev/include_parser_list.md
```diff
@@ -1,6 +1,7 @@
 | OS Name | Parser Class |
 | ---------- | ------ |
 | arista_eos | netutils.config.parser.EOSConfigParser |
+| aruba_aoscx | netutils.config.parser.ArubaConfigParser |
```
itdependsnetworks, line 4: Does this map up to one of the netmiko names? I would try and be consistent with that if possible https://github.com/networktocode/netutils/blob/4a26bcb57b4f71f616de258b713df7333b582409/netutils/lib_mapper.py#L14-L16

netutils/config/parser.py
```diff
@@ -1184,3 +1184,88 @@ class NetscalerConfigParser(BaseSpaceConfigParser):
...
+        """
+        config_lines = []
+        # Grab the banner lines from the config.
+        # Remove the comments from the config.
+        # Add the banner back to the config after the "hostname" command.  The banner ALWAYS comes after the hostname.
+        pattern = r"^banner motd ![\s\S]+?!$|^banner exec ![\s\S]+?!$"
+        banner = re.findall(pattern, config, re.MULTILINE)
+        config = re.sub(pattern, "", config, flags=re.MULTILINE)
+        for line in config.splitlines():
+            if line and not self.is_comment(line) and not line.lstrip().startswith("Current configuration:"):
```
itdependsnetworks, line 1249: I am not sure about removing `Current configuration` from the list. We have been able to get around this on all other OS's, would need to understand why it is required here.

netutils/config/parser.py
```diff
@@ -1184,3 +1184,88 @@ class NetscalerConfigParser(BaseSpaceConfigParser):
...
+                config_lines.append(line.rstrip())
+            if line.lstrip().startswith("hostname"):
```
itdependsnetworks, line 1251: What is this doing?

netutils/config/parser.py
```diff
@@ -1184,3 +1184,88 @@ class NetscalerConfigParser(BaseSpaceConfigParser):
...
+
+        Returns:
+            The non-space and non-comment lines from ``config``.
+        """
+        if self._config is None:
+            config_lines = []
+            for line in self.config.splitlines():
+                if line and not line.isspace():
+                    config_lines.append(line.rstrip())
+            self._config = "\n".join(config_lines)
```
itdependsnetworks, line 1269: Why assign it here, if just going to reassign it in next line? I think this should work:

self._config = self._parse_out_comments("\n".join(config_lines))

docs/dev/include_parser_list.md
lukebates123: Unfortunately that's for the older Aruba OS switches.  Netmiko currently does not have support for aruba aoscx, however Scrapli does, should I add this to the scrapli lib mapper?

netutils/config/parser.py
lukebates123: Aruba uses `!` as both a comment and the banner delimiter.  This is parsing the banners from the config and assigning it to `banner`, remove all the comments, then inject the banner back in after the `hostname` line.

docs/dev/include_parser_list.md
itdependsnetworks: yea, that would be good

netutils/config/parser.py
itdependsnetworks: So you are saying that the banner always happens right after the hostname configuration?

lukebates123: yes, that's correct.

itdependsnetworks: I do **not** think that is a reasonable assumption to make. There may be other (less used) configurations that would come in between.

docs/user/lib_mapper/scrapli.md
```diff
@@ -1,5 +1,6 @@
 | NORMALIZED | | SCRAPLI |
 | ---------- | -- | ------ |
+| aruba_aoscx | → | aruba_aoscx |
```
itdependsnetworks, line 3: Sorry, there is confusion. This is an artifact, not the source. You have to update the actual mappers and then generate this.

netutils/config/parser.py
```diff
@@ -1184,3 +1184,88 @@ class NetscalerConfigParser(BaseSpaceConfigParser):
...
+            if line and not self.is_comment(line) and not line.lstrip().startswith("Current configuration:"):
+                config_lines.append(line.rstrip())
+            if line.lstrip().startswith("hostname"):
+                for banner_line in banner:
+                    config_lines.append(banner_line)
+        config = "\n".join(config_lines)
+        return config
+
+    @property
+    def config_lines_only(self) -> str:
```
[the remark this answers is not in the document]
jeffkala: Great, thanks for the validation. You can disregard this comment now.

netutils/config/parser.py
```diff
@@ -1184,3 +1184,92 @@ class NetscalerConfigParser(BaseSpaceConfigParser):
...
+    def _parse_out_comments(self, config: str) -> str:
+        """Remove comments while retaining the banner end.
+
+        Args:
+            config (str): full config as a string.
+
+        Returns:
+            The non-comment lines from ``config``.
+        """
+        # Aruba AOS-CX uses "!" as both comments and the banner delimiter.
```
jeffkala, line 1241:
```suggestion
        # Aruba AOS-CX uses "!" as both comments and the banner delimiter.
        # Even if another delimiter is used while creating the banner, show run changes the delimiter to use "!".
```

netutils/config/compliance.py
```diff
@@ -277,8 +278,8 @@ def diff_network_config(compare_config: str, base_config: str, network_os: str)
             for parent in line.parents:
                 if parent not in needed_lines:
                     needed_lines.append(parent)
-            needed_lines.append(line.config_line)
-
+            if line.config_line:
```
itdependsnetworks, line 281: What use case and associated test is this accounting for?

netutils/lib_mapper.py
```diff
@@ -247,6 +252,7 @@
     "pluribus": "community.network.netvisor",
     "ruckus_icx": "community.network.icx",
     "vyos": "vyos.vyos.vyos",
+    "aruba_aoscx": "arubanetworks.aoscx",
```
itdependsnetworks, line 255: Sorry, I may have caused some confusion. I thought you were asking about adding `aruba_aoscx` to libraries where they existed, but this does not exist in pyntc as an example. This mapper is intended to map between what a given library calls it.

lukebates123: All the mappings should exist. 
Ansible: https://galaxy.ansible.com/arubanetworks/aoscx
Napalm: https://github.com/napalm-automation-community/napalm-aruba-cx
Scrapli: https://github.com/scrapli/scrapli_community/tree/main/scrapli_community/aruba/aoscx

netutils/config/compliance.py
lukebates123: If there are two banners (`motd` & `exec`) in either the intended or backup config https://github.com/scrapli/scrapli_community/tree/main/scrapli_community/aruba/aoscx but only one in the other config https://github.com/networktocode/netutils/pull/172/files#diff-3df9a197865dd3631c2da5ff59f624886dd2bf16ed16d7fe746b25c1d126f566R2-R4 This will throw a TypeError 
```
Traceback (most recent call last):
  File "/netutils/./test.test", line 68, in <module>
    comply = compliance(features, running_config, generated_config, "aruba_aoscx", "string")
  File "/netutils/netutils/config/compliance.py", line 188, in compliance
    compliance_results.update({feature["name"]: feature_compliance(feature, backup_str, intended_str, network_os)})
  File "/netutils/netutils/config/compliance.py", line 338, in feature_compliance
    feature_data.update(_check_configs_differences(intended_cfg, backup_cfg, network_os))
  File "/netutils/netutils/config/compliance.py", line 66, in _check_configs_differences
    extra = diff_network_config(actual_cfg, intended_cfg, network_os)
  File "/netutils/netutils/config/compliance.py", line 283, in diff_network_config
    return "\n".join(needed_lines)
TypeError: sequence item 4: expected str instance, NoneType found
```

itdependsnetworks:
@lukebates123 great stuff! I/we need some time to review the code but it looks like you have done a great job. Approved GitHub actions so you can see if tests fail or not via CI now. (always run locally of course)

jeffkala (approved):

itdependsnetworks (approved):

PR #177, opened 2022-12-07
Add IOS-XR Parser

### LINKED ISSUE #26: IOS-XR Parser Support
Environment
Python version: 3.8
Nautobot version: 1.0.1
nautobot-golden-config version: develop
Proposed Functionality
Add a parser for IOSXR

Use Case
Already seen it be requested a few times in public chat

jdrew82:
This PR is for adding a parser for the IOS-XR platform. It includes initial tests of the parser and compliance. I'll work to get the other compliance tests in place. Closes #26.


itdependsnetworks:
I think I am good with this, well done @jdrew82!!

Things I was looking for:
* Banner 
* Multiple Banners
* Nested configs with proper parents structure
* Multiple Nested configs with proper parents structure
* Compliance tests

itdependsnetworks (approved):

jeffkala (approved):
approving, but wondering if there is any gotchas with XRs that have admin mode

jeffkala:
> approving, but wondering if there is any gotchas with XRs that have admin mode

May actually be a bad point, I'm trying to find an example of a config to see if there is any differences from the running configuration perspective.

PR #181, opened 2022-12-14
fix: Fixes incorrect behavior when calling tcp_ping with timeout >= 3

### LINKED ISSUE #175: tcp_ping calls socket.shutdown() incorrectly
### Environment
* Python version:  3.8.11
* netutils version:  1.3.0

### Expected Behavior
tcp_ping should return True for timeout values greater than 2

### Observed Behavior
tcp_ping returns False when timeout is 3 or more

### Steps to Reproduce
1. tcp_ping("8.8.8.8", 80, 3)

### Proposed Fix
Change `sckt.shutdown(int(timeout))` to `sckt.shutdown(socket.SHUT_RDWR)`

### Reference
socket.shutdown() accepts a how argument:
[https://docs.python.org/3/library/socket.html#socket.socket.shutdown](url)

abates:
Fixes #175

### DIFF
--- a/netutils/ping.py
+++ b/netutils/ping.py
@@ -25,7 +25,10 @@ def tcp_ping(ip: str, port: int, timeout: int = 1) -> bool:  # pylint: disable=i
     sckt.settimeout(int(timeout))
     try:
         sckt.connect((ip, int(port)))  # pylint: disable=invalid-name
-        sckt.shutdown(int(timeout))
+        sckt.shutdown(socket.SHUT_RDWR)
         return True
-    except socket.error:
+    # We really only want to know if the TCP connection timed out
+    # or if there was an error resolving the DNS name, if anything
+    # else has happened the error should be raised.
+    except (socket.timeout, socket.gaierror):
         return False
--- a/tests/unit/test_ping.py
+++ b/tests/unit/test_ping.py
@@ -7,10 +7,11 @@ from netutils import ping
 ping_data = [
     {"sent": {"ip": "1.1.1.1", "port": 443}, "received": True},
     {"sent": {"ip": "nevergonnagiveyouup.pizza", "port": 443}, "received": False},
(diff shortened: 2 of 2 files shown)


netutils/ping.py
```diff
@@ -25,7 +25,10 @@ def tcp_ping(ip: str, port: int, timeout: int = 1) -> bool:  # pylint: disable=i
...
     try:
         sckt.connect((ip, int(port)))  # pylint: disable=invalid-name
-        sckt.shutdown(int(timeout))
+        sckt.shutdown(socket.SHUT_RDWR)
         return True
-    except socket.error:
+    # We really only want to know if the TCP connection timed out
+    # or if there was an error resolving the DNS name, if anything
+    # else has happened the error should be raised.
+    except (socket.timeout, socket.gaierror):
```
itdependsnetworks, line 33: Please add the finally on this to always disconnect

abates: I tried that, that ends up raising a new exception because if the connection failed you can't actually close it.

jeffkala: wondering if we could save ourselves a bit and use `connect_ex()` instead? https://docs.python.org/3.7/library/socket.html#socket.socket.connect_ex

abates: I'm not sure if that changes the logic required, though. You still end up checking of errno is zero, which would be the same as not catching the exception.

jeffkala: good point

itdependsnetworks: I initially thought that `socket.gaierror` made sense, but I am second guessing that. I think it is best to raise a stacktrace in that instance, since the error should be different for timeout vs a bad dns name.

abates: okay, `finally` added with `close()` and `socket.gaierror` has been removed.

jeffkala (approved):

jeffkala:
good once test pass.

PR #182, opened 2022-12-16
Add parsers for mrv_optiswitch and extreme_netiron

johanek:


johanek:
> Can you read the part in this section about what tests to build and follow that? https://netutils.readthedocs.io/en/latest/dev/dev_config/?h=parser#new-parsers

Hi, what specifically do I need to do for tests? I added tests in `tests/unit/mock/config/parser/`, I missed that there should also be tests in `tests/unit/mock/config/compliance/` - but those directories have been re-worked recently and don't match the docs you linked. Should I be adding similar tests to those I see in `tests/unit/mock/config/compliance/compliance/{os_name}`?

itdependsnetworks:
Exactly

johanek:
I've added the extra tests

netutils/config/parser.py
```diff
@@ -1375,3 +1375,43 @@ def build_config_relationship(self) -> t.List[ConfigLine]:
...
+
+        Args:
+            config (str): The config text to parse.
+        """
+        super(NetironConfigParser, self).__init__(config)
+
+    @property
+    def banner_end(self) -> str:
+        """Demarcate End of Banner char(s)."""
+        raise NotImplementedError("Netiron platform doesn't have a banner.")
```
jeffkala, line 1417:
```suggestion
        raise NotImplementedError("Extreme Netiron platform doesn't have a banner.")
```

jeffkala (approved):

netutils/config/parser.py
```diff
@@ -1375,3 +1375,43 @@ def build_config_relationship(self) -> t.List[ConfigLine]:
...
+    comment_chars: t.List[str] = ["#", "!"]
+    banner_start: t.List[str] = []
+
+    def __init__(self, config: str):
+        """Create ConfigParser Object.
+
+        Args:
+            config (str): The config text to parse.
+        """
+        super(OptiswitchConfigParser, self).__init__(config)
```
itdependsnetworks, line 1392: I see that we did this on Nokia and Aruba, I think it was just missed. I don't think this is needed.

netutils/config/parser.py
```diff
@@ -1375,3 +1375,43 @@ def build_config_relationship(self) -> t.List[ConfigLine]:
...
+    comment_chars: t.List[str] = ["#", "!"]
+    banner_start: t.List[str] = []
+
+    def __init__(self, config: str):
+        """Create ConfigParser Object.
+
+        Args:
+            config (str): The config text to parse.
+        """
+        super(NetironConfigParser, self).__init__(config)
```
itdependsnetworks, line 1412: I see that we did this on Nokia and Aruba, I think it was just missed. I don't think this is needed.

netutils/config/parser.py
johanek: Thanks, removed both of these sections 👍

itdependsnetworks:
I think the last one from me, can you add some tests where there are multiple parents? 
e.g.
```
interface vlan vif20
 tag 20  
 ports 3 
 ipv6 2001::1/120
 ipv6 link-local
 vrrp    
  enable 
  virtual-router 1
   virtual-ip 2001::1 
   enable
```

johanek:
> I think the last one from me, can you add some tests where there are multiple parents?

I added something, I hope it's what you would looking for!

itdependsnetworks (approved):

PR #197, opened 2023-01-19
Prepend OUI Flatbot step to delete existing branch

qduk:
This prepends a step in the Flatbot OUI workflow to delete the branch used by the workflow.

### DIFF
--- a/.github/workflows/oui_filepull.yml
+++ b/.github/workflows/oui_filepull.yml
@@ -12,7 +12,7 @@ jobs:
       BRANCH_NAME: "OUI_Updates"
     steps:
       - name: "Delete existing branch"
-        run: "git branch -D $BRANCH_NAME"
+        run: "git branch -D $BRANCH_NAME || true"
       # Checkout repo
       - name: "Check out code"
         uses: "actions/checkout@v2"
--- a/.github/workflows/oui_filepull.yml
+++ b/.github/workflows/oui_filepull.yml
@@ -11,6 +11,8 @@ jobs:
     env:
       BRANCH_NAME: "OUI_Updates"
     steps:
+      - name: "Delete existing branch"
+        run: "git branch -D $BRANCH_NAME"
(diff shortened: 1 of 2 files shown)


jeffkala (approved):

.github/workflows/oui_filepull.yml
```diff
@@ -11,6 +11,8 @@ jobs:
     env:
       BRANCH_NAME: "OUI_Updates"
     steps:
+      - name: "Delete existing branch"
+        run: "git branch -D $BRANCH_NAME"
```
jeffkala, line 15:
```suggestion
        run: "git branch -D $BRANCH_NAME || true"
```
@abates mentioned that if the branch happened to not exist this would cause a non 0 exit code which would kill the pipeline.  The `|| true` will make it so that it always returns a 0 exit code (true)

jeffkala (changes requested):

flat_postprocess/oui_postprocess.py
```diff
@@ -24,6 +24,6 @@
         oui_mappings.write("# pylint: disable=too-many-lines\n")
         oui_mappings.write("import typing\n\n")
         oui_mappings.write("OUI_MAPPINGS: typing.Dict[str, str] = {\n")
-        for mac, company in OUI_MAPPINGS.items():
+        for mac, company in sorted(OUI_MAPPINGS.items()):
```
qduk, line 27: @jeffkala So I wasn't doing any ordering. I guess I assumed what I received from the web would be consistent in terms of placement. This should order the `OUI_MAPPINGS` dict before writing to the file in the repo.  I'm guessing there will be a lot of changes on the initial PR but after that it should just notice actual additions and removals from the data. Thoughts?

jeffkala: this works for me. Doesn't have to be fancy, but this will help for future PRs.

Issue #96, opened 2022-02-04
Add additional decode/encode to netutils.password 


fischa:
### Environment
* netutils version:  future

### Proposed Functionality
Being able to use netutils to encode/decode the reversible "hash" format (type 9) for some passwords on Junos. Similar to the Cisco Type 5 and 7 passwords.

E.g. like this
* ```netutils.password.decrypt_type9(encrypted_password)```
* ```netutils.password.encrypt_type9(password, seed)```

There is already some translations from the original https://metacpan.org/dist/Crypt-Juniper/view/lib/Crypt/Juniper.pm in Python:
* for the decoding part only: https://github.com/mhite/junosdecode (Python 2..)
* for encoding and decoding: https://github.com/zach-bray/juniper-SNMPv3-crypt/ 

Note: When I read the license of the original Perl module it refers to the same as Perl (which is GPL or the "Artistic License"). Not sure why the 2nd author uses MIT license for his port of it to Python.

### Use Case
* Decoding secrets from existing Junos config to use it as cleartext for another flavour of network operating system. E.g. TACACS password
* Encoding cleartext secret to type 9 hash for using in Junos config templates. E.g. SNMPv3, BGP, TACACS, etc.
* Having all tools in a single lib to consume!

If there are questions around this feature request let me know. It's my first contribution to any NetworkToCode repo :)

Maintainers Edit:

List of encode/decode password types:

- [ ] Junos type 9
- [ ] Arista 
- [ ] Cisco type 9

itdependsnetworks:
Same for Arista: https://aristanetworks.force.com/AristaCommunity/s/article/cvp-tip-encrypt-local-username-passwords-with-configlet-builder#Comm_Kna_ka08C0000008S8XQAU_63 

`# openssl passwd -salt openssl rand -base64 3 -1 donttellanyone`

fischa:
The type 9 "hash" is reversible. In the link you have listed the output gives a type 1 hash which should be MD5 if I recall correctly. MD5 is considered insecure but is not that easy to get to the cleartext input. So I'm not sure if the Arista one is a fit for netutils. There are options in passlib to create real hashes too.

PR #216, opened 2023-02-16
Add `encrypt_type9` function to `netutils.password`

alextremblay:
Related to #96


netutils/password.py
```diff
@@ -233,6 +242,79 @@ def encrypt_type7(unencrypted_password: str, salt: t.Optional[int] = None) -> st
...
+    if salt:
+        if len(salt) != 14:
+            raise ValueError("Salt must be 14 characters long.")
+        salt_bytes = salt.encode()
+    else:
+        # salt must always be a 14-byte-long printable string, often includes symbols like $, @, etc.
+        # if we base85-encode 11 bytes, we get a 14-byte printable string
+        salt_bytes = base64.b85encode(os.urandom(11))
+
+    kdf = Scrypt(
```
abates, line 332: why not use python's `hashlib.script` function?

netutils/password.py
```diff
@@ -233,6 +242,79 @@ def encrypt_type7(unencrypted_password: str, salt: t.Optional[int] = None) -> st
     return encrypted_password
 
 
+def _wpa_base64_encode(data: bytes):
```
abates, line 273: I don't think this is necessary, I think the same can be accomplished with a string translation table. I did some googling and found other implementations that seem more clear, such as [this](https://github.com/BrettVerney/ciscoPWDhasher/blob/74cc0989fb3d3c2802b0389811d1c8022a9ea41f/CiscoPWDhasher/__init__.py#L94).

netutils/password.py
alextremblay: TIL pbkdf2 and scrypt are included in hashlib 😅 

that sounds way better than adding a dependency on cryptography

netutils/password.py
alextremblay: wow, that is a MUCH cleaner implementation. I wish I'd found that when i started researching this

alextremblay: Ok, so i tested BrettVarney's base64 encoding implementation, and the hashes it produces do not match those produced by a cisco switch.

```console
! Tested on IOS-XE 17.3
(config)# username test algorithm-type scrypt secret cisco
(config)# do show run | include username
username test secret 9 $9$x2xAAwQ3MBbEnk$JCxr6MnPb.k5ymK72mTypyRJYH5W74ZRvtLTprCj.xQ
```

```python
>>> password.encrypt_type9_b64_translated("cisco", "x2xAAwQ3MBbEnk") 
'$9$x2xAAwQ3MBbEnk$jcxrWmN/BOKVyMkXSM3ypy1j8hV6XU91vtl3prcJOx0' # does not match cisco's show run
>>> password.encrypt_type9("cisco", "x2xAAwQ3MBbEnk")
'$9$x2xAAwQ3MBbEnk$JCxr6MnPb.k5ymK72mTypyRJYH5W74ZRvtLTprCj.xQ' # DOES match
```

alextremblay: which confirms my prior research. Cisco implements the same bizarre base64 iterated encoding algorithm used by WPA to encode its passwords

netutils/password.py
alextremblay: This change has been made

netutils/password.py
abates: That's interesting, when I run this locally, it generates the same hash as in your IOS-XE example:

```python
#!/usr/bin/env python3

import hashlib
import base64

std_b64chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
cisco_b64chars = "./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
b64table = str.maketrans(std_b64chars, cisco_b64chars)

def encrypt(password, salt):
    password_hash = hashlib.scrypt(password.encode(), salt=salt.encode(), n=16384, r=1, p=1, dklen=32)
    return base64.b64encode(password_hash).decode().translate(b64table)[:-1]

print(encrypt("cisco", "x2xAAwQ3MBbEnk"))
# JCxr6MnPb.k5ymK72mTypyRJYH5W74ZRvtLTprCj.xQ
```

itdependsnetworks: Some oddities with git, but just note @abates other comment about macos. It seems his solution should work, and there is a workaround for MacOS (basically to fail gracefully).

alextremblay: @abates that is indeed interesting. your  code example is identical to the one i wrote to test base64 translation, except that my base64 translation table looked like this:

```python
str.maketrans(
    ALPHABET + '+/',
    ENCRYPT_TYPE9_ENCODING_CHARS
)
```

alextremblay: Ahhhh, i see the discrepancy. `ALPHABET` as defined in `password.py` is composed of ascii_lowercase + ascii_uppercase + digits, not ascii_uppercase + ascii_lowercase + digits.

Ok, I'll re-implement

alextremblay: reimplemented, and the tests are now passing. this base64 scheme behaves correctly

PR #219, opened 2023-02-17
Fix vlanlist_to_config failing on empty list

itdependsnetworks:


jeffkala (approved):

tests/unit/mock/vlan/to_config/empty_list_received.json
```diff
@@ -0,0 +1 @@
+[]
```
jeffkala, line 1:
```suggestion
[]

```

PR #244, opened 2023-03-21
Mikrotik RouterOS Parser

h4ndzdatm0ld:
- Added `mikrotik_routeros` parser
- Compliance Tests
- Parser Tests


tests/unit/mock/config/compliance/config_section_not_parsed/mikrotik_routeros/routeros_feature.py
```diff
@@ -0,0 +1,6 @@
+features = [
```
itdependsnetworks, line 1: Always like to have an example of a banner being parsed.

h4ndzdatm0ld: Hey Ken! The comments at the top of the config files, is that considered a banner here, not just comments? There is a `system note` command, but it just outputs like a normal str config line. 

```[admin@R2] > export terse
# mar/22/2023 17:44:57 by RouterOS 6.48.5
# software id =
#
#
#
/interface wireless security-profiles set [ find default=yes ] supplicant-identity=MikroTik
/ip address add address=172.31.255.30/30 interface=ether1 network=172.31.255.28
/ip dhcp-client add disabled=no interface=ether1
/system identity set name=R2
/system note set note="SOME CRAZY NOTE!"
```

itdependsnetworks: ok, perhaps simpleton of me to presume there is a banner :) 

I do think it would be great to have a few options for "note", like how does it handle multi-line (is that supported), special characters, can you have a double quote in there, etc.. Think of ways to break it

h4ndzdatm0ld: @itdependsnetworks I added to the `parsing` tests to capture the lines that are children of the `/system note set note=..` line. This is the exact input that went into the device is below. If a user inputs a string without adding `\n` the config export will handle it and add it for you, same with special chars.  Device will just escape them when exporting config.

```bash
This is a \"System Note\" for a Mikrotik router.\n\
It includes double quotes (\") and special characters such as:\n\
@, #, $, %, ^, &, *, (, ), _, +, [, ], {, }, |, ;, ',', ., /, <, >, and ?.\n\
\n\
Remember to escape any special characters with a backslash (\\) when necessary.\n\
\n\
This is a multiline note with several lines of text.
```

and when doing a `terse` config export, the following output appears ->

```bash
[admin@ag1.123site.nwk.nj] /system note> export terse
# apr/01/2023 17:16:02 by RouterOS 6.48.5
# software id =
#
#
#
/system note set note="This is a \\\"System Note\\\" for a Mikrotik router.\\n\\\
    \nIt includes double quotes (\\\") and special characters such as:\\n\\\
    \n@, #, \$, %, ^, &, *, (, ), _, +, [, ], {, }, |, ;, ',', ., /, <, >, and \?.\\n\\\
    \n\\n\\\
    \nRemember to escape any special characters with a backslash (\\\\) when necessary.\\n\\\
    \n\\n\\\
    \nThis is a multiline note with several lines of text.\\n\\\
    \nWow, what a great example of a note to ensure proper parsing by NetUtils!"
[admin@ag1.123site.nwk.nj] /system note>
```

hope this helps and captures what you asked for, LMK.

tests/unit/mock/config/parser/base/mikrotik_routeros/routeros_full_received.py
```diff
@@ -0,0 +1,151 @@
...
+    ConfigLine(config_line="/system clock set time-zone-name=America/New_York", parents=()),
+    ConfigLine(config_line="/system identity set name=ag1.123site.nwk.nj", parents=()),
+    ConfigLine(config_line="/system logging add action=remote topics=error", parents=()),
+    ConfigLine(config_line="/system logging add action=remote topics=info", parents=()),
+    ConfigLine(
+        config_line='/system note set note="This is a \\\\\\"System Note\\\\\\" for a Mikrotik router.\\\\n\\\\\\',
+        parents=(),
+    ),
+    ConfigLine(
+        config_line='    \\nIt includes double quotes (\\\\\\") and special characters such as:\\\\n\\\\\\',
```
itdependsnetworks, line 124: In my mind, this line and the below should all be a single line without a parent. As an example I would not see the other lines as children of this parent.

look at: https://github.com/networktocode/netutils/blob/1312c08d58a74ebb63f36ea35d03bec7747628d0/tests/unit/mock/config/parser/base/cisco_ios/ios_full_sent.txt#L288 and https://github.com/networktocode/netutils/blob/1312c08d58a74ebb63f36ea35d03bec7747628d0/tests/unit/mock/config/parser/base/cisco_ios/ios_full_received.py#LL360C34-L360C86

jeffkala: Agreed, I'd expect this to be one single line.

h4ndzdatm0ld: Right on, the more I looked at this, it makes sense to treat this like a banner since it does allow to display upon logging into the device. Made some changes, lmk what ya'll think.  Single line.

itdependsnetworks (approved):

PR #252, opened 2023-03-27
Helper function that returns the proper upgrade path for PANOS firewall firmware updates.

boabdilperez:
* New helper function in lib_helpers to return a list of code versions to form the proper upgrade path during PANOS firewall upgrades.
* Updated documentation, testing, and jinja function mapper.


itdependsnetworks:
I think that this is more generic, would be great to show this for Cisco IOS's or similar and remove anything that is specfic to panos.

netutils/lib_helpers.py
```diff
@@ -13,6 +14,31 @@
 else:
     HAS_NAPALM = True
 
+UPGRADE_PATHS = {
```
jeffkala, line 17: Do we want to keep track of this or should this be up to each user to track?  Seem like it'd require a lot of maintenance.

tests/unit/test_lib_helpers.py
```diff
@@ -2,11 +2,42 @@
...
 def test_get_napalm_getters_napalm_not_installed():
     with pytest.raises(ImportError) as exc:
         get_napalm_getters()
     assert "Napalm must be install for this function to operate." == str(exc.value)
+
+
+def test_get_upgrade_path():
+    assert get_upgrade_path(
+        "9.1.6", "10.1.9", UPGRADE_PATHS["PANOS_FIRMWARE_LIST_OFFICIAL"]
+    ) == [  # pylint: disable=duplicate-code
```
jeffkala, line 22: Probably shouldn't be running pylint on `test` files. Another option here is to change this list to one line (not sure if black formatted it like this) but that would get rid of the dup code error.

itdependsnetworks:
@boabdilperez could you look at #261 as a replacement to this PR?

tests/unit/test_lib_helpers.py
boabdilperez: Black formatted it like this, and just running the test suite from invoke hit this file with this error. If pylint shouldn't be run on test files, is there some way to exempt test_*.py files in the pylint config?

netutils/lib_helpers.py
boabdilperez: You're probably right about this. The function works entirely with externally provided lists of version numbers, and these were convenient extras for the test cases I was building this around.

itdependsnetworks:
To avoid confusion, going to close this in favor of #261, can re-open if it makes sense.

PR #260, opened 2023-04-12
Adds address_type Function

jvanderaa:
Adds in an address_type function that will read a string and return IPv4, IPv6 or raise an IP address error if a bad string is sent in.


jeffkala:
Does the existing not meet the goal you need?
```python
>>> from netutils.ip import ipaddress_address
>>> ipaddress_address('192.0.2.10', 'version')
4
```

qduk:
Closing as the functionality already exists from IPAddress

PR #262, opened 2023-04-14
Palo alto networks

bile0026:
Add parsing for palo alto networks configurations in brace format.


tests/unit/mock/config/parser/base/paloalto_panos/panos_full_received.py
```diff
@@ -0,0 +1,3202 @@
...
+        config_line="        permissions {", parents=("config {", "  mgt-config {", "    users {", "      admin {")
+    ),
+    ConfigLine(
+        config_line="          role-based {",
+        parents=("config {", "  mgt-config {", "    users {", "      admin {", "        permissions {"),
+    ),
+    ConfigLine(
+        config_line="            superuser yes;",
+        parents=(
+            "config {",
```
itdependsnetworks, line 19: hmm, shouldn't the line just be 
```
            "config {",
```

and in below lines the parent be the same as above?

bile0026: The unittest passes, so unless that's wrong I don't think so? Unless I'm misunderstanding something? based on the other brace configs. I saw the entire hierarchy was included for each line.

itdependsnetworks: > The unittest passes, so unless that's wrong I don't think so?

How were the unittests created? Did you generate it by hand? If it was ran by code, the unittest only tests that it is what it was originally. 

> based on the other brace configs. I saw the entire hierarchy was included for each line.

Which brace config are you referring to? I only see F5 as brace, which is in the format I was expecting. https://github.com/networktocode/netutils/blob/develop/tests/unit/mock/config/parser/base/bigip_f5/bigip_full_received.py#L726

bile0026: Moving to conversion to set format.

netutils/config/conversion.py
```diff
@@ -0,0 +1,36 @@
+"""Configuraiton conversion methods for different network operating systems."""
+# pylint: disable=no-member,super-with-arguments,invalid-overridden-method,raise-missing-from,invalid-overridden-method,inconsistent-return-statements,super-with-arguments,redefined-argument-from-local,no-else-break,useless-super-delegation,too-many-lines
```
itdependsnetworks, line 2: Why are there set?

netutils/config/conversion.py
```diff
@@ -0,0 +1,36 @@
...
+
+from typing import Generator, List
+
+
+def paloalto_panos_brace_to_set(config: Generator[str, None, None]) -> List[str]:
```
itdependsnetworks, line 7: We should have tests for these. You should follow the pattern of glob shown here: https://github.com/networktocode/netutils/blob/develop/tests/unit/test_compliance.py#L19-L27

netutils/config/parser.py
```diff
@@ -1419,3 +1421,69 @@ def _build_banner(self, config_line: str) -> t.Optional[str]:
...
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig setting config rematch yes", parents=()),
+            ... ]
+            True
+        """
+        # if config is in palo/json format, convert to set
+        if self.config_lines_only.startswith("config {") and self.config_lines_only is not None:
+            converted_config = paloalto_panos_brace_to_set(self.generator_config)
+            self.generator_config = (line for line in converted_config)
+
+            print(f"Converted Config: {converted_config}")
```
itdependsnetworks, line 1462: remove print.

netutils/config/parser.py
```diff
@@ -1419,3 +1421,69 @@ def _build_banner(self, config_line: str) -> t.Optional[str]:
...
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig system hostname firewall1", parents=()),
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig system panorama local-panorama panorama-server 10.0.0.1", parents=()),
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig system panorama local-panorama panorama-server-2 10.0.0.2", parents=()),
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig setting config rematch yes", parents=()),
+            ... ]
+            True
+        """
+        # if config is in palo/json format, convert to set
+        if self.config_lines_only.startswith("config {") and self.config_lines_only is not None:
+            converted_config = paloalto_panos_brace_to_set(self.generator_config)
```
itdependsnetworks, line 1459: If we are going to normalize to set, why is this based on a brace parser?

bile0026: I'm not following? This would allow panos backups to be in brace or set format, but the class is inheriting from basespaceparser. Set is preferred since, as discussed, brace would require custom compliance methods.

netutils/config/conversion.py
bile0026: Test added

tests/unit/mock/config/parser/base/paloalto_panos/panos_full_received.py
```diff
@@ -0,0 +1,381 @@
...
+    ConfigLine(
+        config_line="set devices localhost.localdomain deviceconfig system service disable-telnet yes", parents=()
+    ),
+    ConfigLine(
+        config_line="set devices localhost.localdomain deviceconfig system service disable-http yes", parents=()
+    ),
+    ConfigLine(config_line="set devices localhost.localdomain deviceconfig system hostname firewall1", parents=()),
+    ConfigLine(config_line='set devices localhost.localdomain deviceconfig system login-banner "', parents=()),
+    ConfigLine(
+        config_line=" ************************************************************************",
```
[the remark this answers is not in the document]
bile0026: Banner adjusted

netutils/config/parser.py
itdependsnetworks: as we spoke on slack, my mistake, I got confused with Space vs Brace.

netutils/config/parser.py
```diff
@@ -1419,3 +1421,99 @@ def _build_banner(self, config_line: str) -> t.Optional[str]:
...
+            ... [
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig system hostname firewall1", parents=()),
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig system panorama local-panorama panorama-server 10.0.0.1", parents=()),
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig system panorama local-panorama panorama-server-2 10.0.0.2", parents=()),
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig setting config rematch yes", parents=()),
+            ... ]
+            True
+        """
+        # if config is in palo/json format, convert to set
+        if self.config_lines_only.startswith("config {") and self.config_lines_only is not None:
```
itdependsnetworks, line 1490: My gut says this should be a bit more resilient, such as if the second line has "config {" or something similar.

tests/unit/mock/config/parser/base/paloalto_panos/panos_full_received.py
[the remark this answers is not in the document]
itdependsnetworks: I still see an issue. Everything should be under a single parent for the banner, the parent in this case would be `set devices localhost.localdomain deviceconfig system login-banner "`

itdependsnetworks:
getting a pylint issue now.

netutils/config/parser.py
bile0026: Modifying this to check for any lines in the config that end in { would trigger conversion

tests/unit/mock/config/compliance/section/paloalto_panos/panos_full_sent.txt
```diff
@@ -0,0 +1,20 @@
+set mgt-config users admin phash *
+set mgt-config users admin permissions role-based superuser yes
+set mgt-config users admin public-key thisisasuperduperlongbase64encodedstring
+set mgt-config users panadmin permissions role-based superuser yes
+set mgt-config users panadmin phash passwordhash
+set devices localhost.localdomain deviceconfig system hostname firewall1
+set devices localhost.localdomain deviceconfig system login-banner
```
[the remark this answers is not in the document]
bile0026: yes there is, fixing this

netutils/config/parser.py
bile0026: Modifying this to check for any lines in the config that end in { would trigger conversion

pyproject.toml
```diff
@@ -27,6 +27,7 @@ include = [
 [tool.poetry.dependencies]
 python = "^3.7"
 napalm = {version = "^4.0.0", optional = true}
+pan-python = "^0.22.0"
```
itdependsnetworks, line 30: I don't see where this is used? 

Also, we would not support this as a required dependency.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "netutils"
-version = "1.4.1"
+version = "1.4.2"
```
itdependsnetworks, line 3: You can remove this, we will take care of it and being that it is a new feature, it would go into 1.5.0.

pyproject.toml
bile0026: i will remove.

pyproject.toml
bile0026: This was to test getting this to deploy in a customer environment. I'll remove it before the final push.

tests/unit/mock/config/parser/base/paloalto_panos/panos_full_received.py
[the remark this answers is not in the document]
bile0026: okay I think I just about have this figured out now. Will be pushing changes in a bit.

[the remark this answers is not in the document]
bile0026: I think this should be as desired now with the latest push.

netutils/config/conversion.py
```diff
@@ -0,0 +1,34 @@
+"""Configuraiton conversion methods for different network operating systems."""
+
+import typing as t
+
+
+def paloalto_panos_brace_to_set(config: t.Generator[str, None, None]) -> t.List[str]:
+    """Convert Palo Alto Brace/JSON format configuration to set format."""
```
jeffkala, line 7:
```suggestion
    """Convert Palo Alto Brace format configuration to set format."""
```

netutils/config/conversion.py
```diff
@@ -0,0 +1,34 @@
+"""Configuraiton conversion methods for different network operating systems."""
```
qduk, line 1:
```suggestion
"""Configuration conversion methods for different network operating systems."""
```

netutils/config/compliance.py
```diff
@@ -20,9 +20,14 @@
...
     "mrv_optiswitch": parser.OptiswitchConfigParser,
     "extreme_netiron": parser.NetironConfigParser,
+    "paloalto_panos": parser.PaloAltoNetworksConfigParser,
     "mikrotik_routeros": parser.RouterOSConfigParser,
 }
 
+conversion_map: t.Dict[str, t.Type[parser.BaseConfigParser]] = {
+    "paloalto_panos": parser.PaloAltoNetworksConfigParser,
+}
+
```
qduk, line 30: Can we move this to the `conversion.py` file?

tests/unit/test_conversion.py
```diff
@@ -0,0 +1,30 @@
...
+from netutils.config.compliance import conversion_map
+
+MOCK_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "mock", "config", "conversion")
+
+TXT_FILE = "_sent.txt"
+CONVERTED_FILE = "_converted.txt"
+
+conversion_files = []
+
+for network_os in list(conversion_map.keys()):
```
itdependsnetworks, line 18: Jeff pointed out that `config/conversion/` is defined both in the `MOCK_DIR` and in the `glob` line of code. Likely it should just be: 

```python
    for _file in glob.glob(f"{MOCK_DIR}/{network_os}/*{TXT_FILE}"):
        conversion_files.append([_file, network_os])
```


```
tests/unit/test_conversion.py::test_config_conversion[_file0] SKIPPED (got empty parameter set ['_file'], function test_config_conversion at /local/tests/unit/test_conversion.py:22) [ 34%]
```

netutils/config/parser.py
```diff
@@ -1419,3 +1421,80 @@ def _build_banner(self, config_line: str) -> t.Optional[str]:
...
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig system panorama local-panorama panorama-server 10.0.0.1", parents=()),
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig system panorama local-panorama panorama-server-2 10.0.0.2", parents=()),
+            ...     ConfigLine(config_line="set devices localhost.localdomain deviceconfig setting config rematch yes", parents=()),
+            ... ]
+            True
+        """
+        # assume configuration does not need conversion
+        _needs_conversion = False
+
+        # if config is in palo/json format, convert to set
```
jeffkala, line 1466:
```suggestion
        # if config is in palo brace format, convert to set
```

netutils/config/compliance.py
itdependsnetworks, line 30: This should be:

* Move to conversion.py
* Map to `'paloalto_panos_brace_to_set': paloalto_panos_brace_to_set`
* Point the test to ^^

tests/unit/mock/config/conversion/paloalto_panos/paloalto_basic_converted.txt
```diff
@@ -0,0 +1,20 @@
+set mgt-config users admin phash *
```
jeffkala, line 1: After a conversion was done, was this config actually pasted onto a true palo alto device.  Just to make sure nothing is converting wrong?

bile0026: I do not have one to test on. If someone has one that would be good. I did a visual compare to the original configs they were based on and they looked good.

tests/unit/test_conversion.py
```diff
@@ -0,0 +1,37 @@
...
+
+    truncate_file = os.path.join(MOCK_DIR, _file[0][: -len(TXT_FILE)])
+
+    sent_cfg = get_text_data(os.path.join(MOCK_DIR, _file[0]))
+    line_cfg = (line for line in sent_cfg.split("\n "))
+    converted_cfg = paloalto_panos_brace_to_set(line_cfg)
+    for i, _line in enumerate(converted_cfg):
+        string_cfg += _line
+        if i < len(converted_cfg) - 1:
+            string_cfg += "\n"
```
itdependsnetworks, line 35: I am not sure i understand what this this doing, but perhaps has to do with converting a text file to a list of strings that represent lines??

Normally with something like that meant:
* config.splitlines()
* '\n'.join(config)

depending on the direction.

This does beg the question, what does and should the conversion take and return? e.g. a string or a list?

tests/unit/mock/config/conversion/paloalto_panos/paloalto_basic_converted.txt
jeffkala: for what its worth, batfish parsing is complianing about the banner quouting.
```
>>> issues = bf.q.initIssues().answer().frame().iloc[65]
>>> issues
Nodes                                      None
Source_Lines                 [configs/p.cfg:[]]
Type                    Parse warning (redflag)
Details           Improperly-quoted string: "\n
Line_Text                                  None
Parser_Context                             None
Name: 65, dtype: object
```

bile0026: The banner in the converted config is identical to how it shows up in a backup file. As far as I know there is no extra tweaking going on with the backup files, they should be raw from the device/panorama. Can batfish look at a brace format config and does it give the same error?

pyproject.toml
itdependsnetworks: I do not understand why the poetry lock file is changed if this is not needed?

tests/unit/mock/config/conversion/paloalto_panos/paloalto_basic_converted.txt
bile0026: So upon further investigation, it does not appear that you can set a multi-line banner from the CLI (only from the GUI), but the format included is how the backup/set format presents it. Kind of seems like this would be the proper format to keep it in since that's also how golden config is going to compare it?

pyproject.toml
bile0026: Sorry, I'm in the habit of running poetry update. I'll get this reverted.

PR #286, opened 2023-06-09
Move password helper names to be namespaced with vendor information. …

itdependsnetworks:
…Assign new name to old name to not break semver.


netutils/password.py
```diff
@@ -454,3 +454,11 @@ def encrypt_juniper(unencrypted_password: str, salt: t.Optional[int] = None) ->
             encrypted_password += new_character
 
     return encrypted_password
+
+
+# Provide until transition to 2.0
+compare_type5 = compare_cisco_type5
```
jeffkala, line 460: Should we print/log a deprecation warning? Or just notate it properly with 2.0 drops and we depredate it?

jeffkala (approved):

netutils/password.py
itdependsnetworks: Yes, my only thoughts on where to put that was here: https://github.com/networktocode/netutils/pull/287/files#diff-fe90e0ca8477ff0667f963638d45d1304773fcd1d1d1b5f0a94eab0cebee4826R12

jeffkala: Looks good to me, I would say this one is a announcment, but would consider the python3.7 support drop to be under a `#Removed` section.

itdependsnetworks: will add as PR's are put in for the two parsers and the 3.7 support the actual PR number.

itdependsnetworks:
@johanek  @abates please note the change in function names.

PR #289, opened 2023-06-09
Closes #288 - added hostname_regex function and jinja filter

### LINKED ISSUE #288: Hostname Regex Filter
### Environment
* netutils version:  

### Proposed Functionality

As a Network Engineer I want jinja filter (test) that I can use to evaluate a regex against a device hostname so that I can parse embedded logic within my hostname standards.  

I'd like a helper function in netutils that takes a hostname string and regex as input and returns a boolean based on the regex match. This would then also be exposed as a jinja filter (test) for use in places like Nautobot.

### Use Case

Often I want to make decisions based on the inherent logic comprised in my device hostnames. Most of the time, this results in boolean evaluation of regexs against my hostnames.

lampwins:
Added a new module `netutils.hostname` with a single function `hostname_regex` that is also registered as a jinja filter.

### DIFF
--- a/docs/dev/code_reference/hostname.md
+++ b/docs/dev/code_reference/hostname.md
@@ -1,0 +1,5 @@
+# Hostname
+
+::: netutils.hostname
+    options:
+        show_submodules: True
--- a/docs/user/include_jinja_list.md
+++ b/docs/user/include_jinja_list.md
@@ -18,6 +18,7 @@
 | section_config | netutils.config.compliance.section_config |
 | fqdn_to_ip | netutils.dns.fqdn_to_ip |
 | is_fqdn_resolvable | netutils.dns.is_fqdn_resolvable |
+| hostname_regex | netutils.hostname.hostname_regex |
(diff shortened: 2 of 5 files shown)


itdependsnetworks:
I like it, it had just come up on public slack. I think it makes sense to have:

regex_search, returns a true/false. 
regex_match, returns a list of the capture group. 

The value I see in these is for Jinja, similar to https://netutils.readthedocs.io/en/latest/user/lib_use_cases_jinja_filters/#ipaddress-convenience-functions, and I am not sure the rich object would work, such as `match[1]` after processed through jinja.

abates:
I'm in agreement with @itdependsnetworks. I think this should be generalized to `regex_match` and `regex_search` since there is nothing inherent with the filter related to hostnames.

itdependsnetworks:
Addressed in #317

itdependsnetworks:
This should be replaced with #318 will likely merge mon/tues to prepare for a 1.5.0 release.

itdependsnetworks:
fixed in #318

PR #339, opened 2023-08-10
Fix for ASN functions

mzbroch:
Use proper names with regards to RFC :

```

   asplain
      refers to a syntax scheme of representing all AS numbers using
      decimal integer notation.  Using asplain notation, an AS number of
      value 65526 would be represented as the string "65526" and an AS
      number of value 65546 would be represented as the string "65546".

   asdot+
      refers to a syntax scheme of representing all AS numbers using a
      notation of two integer values joined by a period character: <high
      order 16-bit value in decimal>.<low order 16-bit value in
      decimal>.  Using asdot+ notation, an AS number of value 65526
      would be represented as the string "0.65526" and an AS number of
      value 65546 would be represented as the string "1.10".

   asdot
      refers to a syntax scheme of representing AS number values less
      than 65536 using asplain notation and representing AS number
      values equal to or greater than 65536 using asdot+ notation.
      Using asdot notation, an AS number of value 65526 would be
      represented as the string "65526" and an AS number of value 65546
      would be represented as the string "1.10".
```


itdependsnetworks:
We would have to deprecate, and wait until 2.0 to release.

itdependsnetworks:
wait, nvm, this hasn't been released yet.

jeffkala:
Validated only `asn_to_int` was the only function released so far. So we should be good.

jeffkala:
Also get pytest passing.

qduk:
@mzbroch Can you revert the table formatting? While it does objectively look better, we believe that's causing the failed test.

itdependsnetworks (approved):

PR #340, opened 2023-08-10
Correct Netscaler parser slug

jdrew82:
This PR is to correct the slug used for Citrix Netscaler to align with netmiko driver that the other platforms use.

The tests have also been updated to match.


jdrew82:
> This would require major version bump correct? As its a breaking change? Maybe we have to keep both for now and have a deprecation warning of sorts for the `citrix_netscaler` one?

This could be considered a breaking change, yes. If we want to keep the original and just have two entries for the same platform I'm fine with that though I wonder how the tests would handle it.

jeffkala:
Think i'm good with this.  I was researching for a decent way to do a deprecation warning on dictionary key, but doesn't seem there's a easy way.

jeffkala (approved):

qduk (approved):

PR #347, opened 2023-08-19
Pin mkdocstrings-python to 1.4.0 to have compatibility with griffe 0.…

itdependsnetworks:
…33.0


itdependsnetworks:
See https://github.com/mkdocstrings/griffe/issues/197 and https://github.com/mkdocstrings/python/discussions/98 for history

PR #374, opened 2023-09-13
[NN20-prep] - netutils/lib_mapper.py

pato23arg:
adds the following platforms to MAIN_LIB:

- ruckus access point
- ruckus smartzone API
- mikrotik routeros API


netutils/lib_mapper.py
```diff
@@ -359,7 +359,10 @@
 _MAIN_LIB_MAPPER["huawei_vrp"] = "huawei_vrp"
 _MAIN_LIB_MAPPER["lenovo_cnos"] = "lenovo_cnos"
 _MAIN_LIB_MAPPER["lenovo_enos"] = "lenovo_enos"
+_MAIN_LIB_MAPPER["mikrotik_routeros_api"] = "mikrotik_routeros_api"
```
itdependsnetworks, line 362: we should not have _api on these. When we use this in nornir_nautobot, it will have to come with a preferred driver (meaning netmiko, napalm, or api) that will make this work. This should just be the defintions of the mappings.

pato23arg: ok so this should not refer to any nautobot platform, just define the OS name.

jeffkala (approved):

jeffkala (approved):

PR #377, opened 2023-09-21
Addition of Wildcard Mask Function

steve-hodgkiss:
Cisco access-list use a wildcard mask rather than a standard mask
Function to create mask in "reverse" format : e.g. 255.255.255.0 > 0.0.0.255

### DIFF
--- a/netutils/ip.py
+++ b/netutils/ip.py
@@ -452,6 +452,27 @@ def cidr_to_netmaskv6(cidr: int) -> str:
         return str(ipaddress.IPv6Address(((1 << cidr) - 1) << (128 - cidr)))
     raise ValueError("Parameter must be an integer between 0 and 128.")
 
+def cidr_to_netwildmask(cidr: int) -> str:
+    """Creates a decimal format of a CIDR value in Cisco wild card mask format used in access-lists.
+
+    **IPv4** only.  No IPv6 equivalent.
+
+    Args:
+        cidr: A CIDR value.
+
+    Returns:
+        Decimal format representation of CIDR value in Cisco wildcard mask format.
+
+    Examples:
+        >>> from netutils.ip import cidr_to_netwildmask
+        >>> cidr_to_netwildmask(24)
+        '0.0.0.255'
+        >>> cidr_to_netwildmask(17)
+        '0.0.127.255'
+    """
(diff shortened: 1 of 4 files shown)


jeffkala:
Thanks for the interest @steve-hodgkiss , we require every function to be unittested in netutils.  Can you get a few tests added.  Should follow the same principles as https://github.com/networktocode/netutils/blob/93597ec3d1a44b8c75a427c92de0d02eab90fd57/tests/unit/test_ip.py#L563

netutils/ip.py
```diff
@@ -453,6 +453,29 @@ def cidr_to_netmaskv6(cidr: int) -> str:
     raise ValueError("Parameter must be an integer between 0 and 128.")
 
 
+def cidr_to_netwildmask(cidr: int) -> str:
```
jeffkala, line 456:
```suggestion
def cidr_to_wildcardmask(cidr: int) -> str:
```

steve-hodgkiss: Function name changed to ``cidr_to_wildcardmask``. 
Added unit tests. 
Updated jinja convenience function list

qduk (approved):
LGTM! Thank you!

jeffkala:
Doing one more review here, I actually think since we extend ipaddress natively this is already available via filters.  I know its available via ipaddress module already.

```python
>>> import ipaddress
>>> a = ipaddress.ip_network("10.0.0.0/10")
>>> a.hostmask
IPv4Address('0.63.255.255')
```

jeffkala:
> Doing one more review here, I actually think since we extend ipaddress natively this is already available via filters. I know its available via ipaddress module already.
> 
> ```python
> >>> import ipaddress
> >>> a = ipaddress.ip_network("10.0.0.0/10")
> >>> a.hostmask
> IPv4Address('0.63.255.255')
> ```

Will validate shortly how it can be used with existing filters.

jeffkala:
@steve-hodgkiss validated this should already exist via this `{{ "192.168.0.0/24" | ipaddress_network("hostmask") }}` 

Does this solve the problem or did you still want to pursue this?

steve-hodgkiss:
Fine either way - if you think the function is not needed then happy for this to be closed out.

PR #416, opened 2023-11-21
fix: 🐛 fixes for \n characters in inappropriate places and parsing bug

bile0026:
Fix bad logic with line.startswith.
Add clean function to handle when lines are inappropriately split between double quotes in some backup files.


netutils/config/conversion.py
```diff
@@ -9,6 +10,119 @@
...
+        ...                 transport UDP;
+        ...                 port 514;
+        ...                 format BSD;
+        ...                 server 2.2.2.2;
+        ...                 facility LOG_USER;
+        ...             }
+        ...             }
+        ...             format {
+        ...             config "CEF:0|Palo Alto Networks|PAN-OS|$sender_sw_version|$result|$type|1|rt=$cef-formatted-receive_time deviceExternalId=$serial shost=$host cs3Label=Virtual System cs3=$vsys act=$cmd duser=$admin destinationServiceName=$client msg=$path externalId=$seqno PanOSDGl1=$dg_hier_level_1 PanOSDGl2=$dg_hier_level_2 PanOSDGl3=$dg_hier_level_3 PanOSDGl4=$dg_hier_level_4 PanOSVsysName=$vsys_name dvchost=$device_name PanOSActionFlags=$actionflags
+        ... Optional: cs1Label=Before Change Detail cs1=$before-change-detail cs2Label=After Change Detail cs2=$after-change-detail";
```
itdependsnetworks, line 46: Just a note for review, this is an example of a newline inserted.

bile0026:
> Hey @bile0026. Thanks for the PR. Can you add tests for `palo_alto_clean_newlines` that show the following:
> 
> * When no newline is used.
> * What happens when a match is found? e.g. does it concatenate it to a single line?
> * How does this affect the compliance?

Added tests for the first 2. Yes, it finds the rest of the line up to the `"` and concatenates it to a single line.

This is specific to when there's a `\n` character between quote marks "". These aren't supposed to be there but if a user copy/pastes a config into the GUI it will allow it. If these `\n` characters are in the backup config it breaks compliance rules and never shows up in the Golden Config section because GC ends up treating it like a new config section because it drops it to a new line without indentation.
```text
this is a config line {
    section1 {
        configuration line
drops to a new line with no indent;
        next configuration line;
    }
}   
```

itdependsnetworks (approved):

netutils/utils.py
```diff
@@ -86,6 +86,7 @@
     "uptime_seconds_to_string": "time.uptime_seconds_to_string",
     "uptime_string_to_seconds": "time.uptime_string_to_seconds",
     "get_napalm_getters": "lib_helpers.get_napalm_getters",
+    "palo_alto_clean_newlines": "config.conversion.palo_alto_clean_newlines",
```
jeffkala, line 89: other utils are called `paloalto` vs `palo_alto` would like to standardize.

docs/user/include_jinja_list.md
```diff
@@ -17,6 +17,7 @@
 | feature_compliance | netutils.config.compliance.feature_compliance |
 | find_unordered_cfg_lines | netutils.config.compliance.find_unordered_cfg_lines |
 | section_config | netutils.config.compliance.section_config |
+| palo_alto_clean_newlines | netutils.config.conversion.palo_alto_clean_newlines |
```
jeffkala, line 20: same as before. `paloalto` vs `palo_alto` in the naming.  Should use `paloalto` since that is the pattern already existing.

jeffkala (changes requested):

docs/user/include_jinja_list.md
bile0026: Addressed in latest push

jeffkala (approved):

itdependsnetworks:
@qduk will merge by friday unless you have a concern.

qduk (approved):
LGTM

PR #425, opened 2023-12-12
Add aerleon and capirca platform mappers

scetron:
Added common network platforms supported by Aerleon and Capirca.


netutils/lib_mapper.py
```diff
@@ -350,6 +411,20 @@
 NETMIKO_LIB_MAPPER_REVERSE = copy.deepcopy(NETMIKO_LIB_MAPPER)
 NTCTEMPLATES_LIB_MAPPER_REVERSE = copy.deepcopy(NTCTEMPLATES_LIB_MAPPER)
 
+NETUTILS_LIBMAPPER_MAPPER = {
```
itdependsnetworks, line 414: What does this offer over "MAIN_LIB_MAPPER"?

itdependsnetworks:
Can we just list all of them so they don't trickle in as much?

scetron:
Do you mean list all of the platforms supported by Aerleon/Capirca? I didn't add them only because many of them don't have another name to be mapped to, which I am fine with adding as I see in some other mappers, it's the same.

itdependsnetworks:
Yea, would be good to know what those are, even if commented out for now.

netutils/lib_mapper.py
scetron: It appears to me that main lib mapper is more focused on the end platform, rather than the library that has a platform. I can't do `MAIN_LIB_MAPPER['aerleon']` or `MAIN_LIB_MAPPER['napalm']` unless I am mistaken?

scetron: Any thoughts on this one?

qduk:
@scetron Can you add the new mappers to [this](https://github.com/networktocode/netutils/blob/8bddf77a3b8ee7c86761b23c6bdd6021ce513074/development_scripts.py#L35) dictionary. It should create the docs once you run the script.

netutils/lib_mapper.py
itdependsnetworks: ok, should it be `    "netmiko": list(NETMIKO_LIB_MAPPER.keys()),` then?

itdependsnetworks: Thinking about it more, it should just be

```python
from netutils import lib_mapper

getattr(lib_mapper, f"{lib.upper()}_LIB_MAPPER")
```

itdependsnetworks:
@scetron let's remove the NETUTILS_LIBMAPPER_MAPPER and add to the dictionary, just to get capirca and aerleon in, we can take the topic of NETUTILS_LIBMAPPER_MAPPER to another issue.

scetron:
@itdependsnetworks Sounds good. Removed. @qduk Added to dev scripts and generated the new docs.

qduk (approved):
Thanks @scetron!

itdependsnetworks (approved):

PR #429, opened 2023-12-19
Updates parsers for issues 414, 415, and 422

qduk:
1. Reverse `huawei` as we had it incorrect. [Example usage of huawei network driver](https://github.com/napalm-automation-community/napalm-huawei-vrp?tab=readme-ov-file#quick-start)
2. Adds mapping for `cisco_xe` where applicable.
3. Updates normalized of `cisco_xr` per [this](https://github.com/ktbyers/netmiko/blob/develop/netmiko/cisco/cisco_xr.py)

Because `cisco_xe` maps to the same platform in some reverse mappings, I popped that key out for the reverse to non-reverse mapping test. I'd definitely the review on this as I want to make sure I'm not missing anything. (Plus these mappers are always a little confusing.)


docs/user/lib_mapper/netutilsparser.md
```diff
@@ -6,7 +6,7 @@
 | cisco_aireos | → | cisco_aireos |
 | cisco_asa | → | cisco_asa |
 | cisco_ios | → | cisco_ios |
-| cisco_iosxr | → | cisco_iosxr |
+| cisco_iosxr | → | cisco_xr |
```
itdependsnetworks, line 9: isn't it iosxr? https://github.com/networktocode/netutils/blob/e4363774b99539181a02d8d19bdf37d42c03fb70/netutils/config/compliance.py#L16

qduk: The normalized (netmiko) doesn't match `cisco_iosxr` it matches `cisco_xr`

itdependsnetworks (approved):

PR #464, opened 2024-02-20
Netutils/acl mod

mzbroch:


netutils/acl.py
```diff
@@ -513,9 +570,9 @@ def match(self, rule: ACLRule) -> str:
             The response from the rule that matched, or `deny` by default.
         """
         for item in self.rules:
-            if item.match(self.class_obj(rule)):
-                return str(item.action)
-        return str(item.deny)  # pylint: disable=undefined-loop-variable
+            if item.match(rule):  # mzb: bugfix
+                return True  # mzb: change to bool
+        return False  # pylint: disable=undefined-loop-variable
```
itdependsnetworks, line 575: I had this as string as most vendors have more than just permit or deny, this is Palo's

```
allow
deny
drop
reset client
reset server
reset client and server
```

mzbroch: IMO `match` method should return only True/False if the rule is/not matched by the check. This should only check if A contains B.

itdependsnetworks: I think it is reasonable to have different levels.

netutils/acl.py
```diff
@@ -147,30 +151,54 @@ def _check_schema(data: t.Any, schema: t.Any, verify: bool) -> None:
             raise ValueError()
 
 
+def get_attributes(obj):
```
itdependsnetworks, line 154: lets make private.

netutils/acl.py
```diff
@@ -179,55 +207,80 @@ def __init__(self, data: t.Any, *args: t.Any, **kwargs: t.Any):  # pylint: disab
         Examples:
             >>> from netutils.acl import ACLRule
             >>>
-            >>> acl_data = dict(
+            >>> rule = ACLRule(
```
itdependsnetworks, line 210: I think we need a loader of sorts, something where you can just through it a dictionary, and it handles everything else.

netutils/acl.py
```diff
@@ -179,55 +207,80 @@ def __init__(self, data: t.Any, *args: t.Any, **kwargs: t.Any):  # pylint: disab
...
             setattr(self, attr, _attr_data)
+
         self.result_data_check()
         self.validate()
-        self.expanded_rules = _cartesian_product(self.processed)
-        if self.filter_same_ip:
-            self.expanded_rules = [item for item in self.expanded_rules if item["dst_ip"] != item["src_ip"]]
+
+    @property
+    def _expanded_rules(self):
```
itdependsnetworks, line 264: Why make this private? I think there is value in having access to this and perhaps in serialize format as well.

netutils/acl.py
mzbroch: I think, I see this in a little different way :
```
def match_action(action, match_action):
    deny_actions = ['deny', 'drop', 'reset client', 'reset server', ...]
    if action in deny_actions and match_action in deny_actions:
        return True
    else:
        return False
```

I would really like to avoid putting opinionated logic into generic `match` resolution while we provide easily extensible framework

netutils/acl.py
mzbroch: This is now supported :
```
acl_data = {'action': permit, src='', dst='' ... }
ACLRule(**acl_data)
```
Can't think of simpler way ?

netutils/acl.py
mzbroch: Agreed

PR #466, opened 2024-02-23
Improve performance of jinja2_convenience_function by not importing NAPALM when called

glennmatthews:
Discovered this while investigating startup performance of the `nautobot-server` command as a part of nautobot/nautobot#4292. I found that the `netutils.utils.jinja2_convenience_function()` function, which is called as a part of Nautobot's startup sequence, turned out to be surprisingly expensive to call, the reason being that `jinja2_convenience_function()` imports (but does not itself call) `netutils.lib_helpers.get_napalm_getters()`, which in turn automatically imports the entire NAPALM library. 

Importing `napalm` takes anywhere from 0.5 to 1 second all by itself, and we don't actually **need** that library imported until we actually *call* `get_napalm_getters()`. My solution here is to simply move the import inside the `get_napalm_getters()` function so that it runs when the function is called, rather then simply when the function is initially imported. This makes `jinja2_convenience_function()` notably faster to call.

(edit on noticing CI failure - Why are there tests that CI includes that aren't run by `invoke tests`? I ran `invoke tests` successfully locally before pushing this branch up. I'll investigate, but that's a bit annoying. :smile: )

### DIFF
--- a/netutils/lib_helpers.py
+++ b/netutils/lib_helpers.py
@@ -5,14 +5,6 @@ import typing as t
 
 from netutils.lib_mapper import NAPALM_LIB_MAPPER
 
-try:
-    from napalm import get_network_driver
-    from napalm.base.exceptions import ModuleImportError
-except ImportError:
-    HAS_NAPALM = False
-else:
-    HAS_NAPALM = True
-
 
 def get_napalm_getters() -> t.Dict[str, t.Dict[str, bool]]:
     """Utility to return a dictionary of napalm getters based on install napalm version.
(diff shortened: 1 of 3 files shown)


itdependsnetworks (approved):

itdependsnetworks:
The reason is we run optional tests as these are optional dependencies. I am not sure this is the best pattern, but what we have done so far. 

https://github.com/networktocode/netutils/blob/develop/.github/workflows/ci.yml#L207-L210

qduk (approved):

PR #473, opened 2024-03-01
Add Sort List of IPs

jdrew82:
This PR adds a simple method to take a string list of CIDRs and returns them sorted.


tests/unit/test_ip.py
```diff
@@ -617,3 +618,25 @@ def test_ipaddress_network(data):
 @pytest.mark.parametrize("data", IS_CLASSFUL)
 def test_is_classful(data):
     assert ip.is_classful(**data["sent"]) == data["received"]
+
+
+def test_sort_list_cidrs():
+    sent = "10.0.10.0/24, 10.0.100.0/24, 10.0.12.0/24, 10.0.200.0/24"
```
itdependsnetworks, line 624: What is the expectation of `"10.0.0.0/24, 10.0.0.0/16, 10.0.0.0/18"`?

Does it make sense to convert this to `ip_sorted` and leave it on the user to re-add the subnet mask back?

jdrew82: Hmmm, good point... I guess I would just expect it to have the CIDR twice in the returned string.

jdrew82: I would expect it to just return the list of CIDRs in the right order, so should be the /16, /18, and /24 I'd guess, from largest to smallest?

jdrew82: Just added a test validating this assumption.

netutils/ip.py
```diff
@@ -596,3 +597,33 @@ def get_usable_range(ip_network: str) -> str:
...
+    if sort_type not in ["address", "interface", "network"]:
+        raise ValueError("Invalid sort type passed. Must be `address`, `interface`, or `network`.")
+    if isinstance(ips, list):
+        ips_list = ips
+    elif (isinstance(ips, str) and "," not in ips) or not isinstance(ips, str):
+        raise ValueError("Not a concatenated list of IPs as expected.")
+    elif isinstance(ips, str):
+        ips_list = ips.replace(" ", "").split(",")
+
+    functions = {"address": ipaddress.ip_address, "interface": ipaddress.ip_interface, "network": ipaddress.ip_network}
```
Kircheneer, line 621:
```suggestion
    functions: t.Dict[str, Callable] = {"address": ipaddress.ip_address, "interface": ipaddress.ip_interface, "network": ipaddress.ip_network}
```

Maybe try this to resolve the error

jdrew82: Thank you, that's what I was thinking might be needed.

netutils/ip.py
```diff
@@ -596,3 +597,37 @@ def get_usable_range(ip_network: str) -> str:
         lower_bound = str(net[1])
         upper_bound = str(net[-2])
     return f"{lower_bound} - {upper_bound}"
+
+
+def sort_list_ips(ips: t.Union[str, t.List[str]], sort_type: str = "network") -> str:
+    """Given a concatenated list of CIDRs sorts them into the correct order and returns as concatenated string.
```
qduk, line 603: Can you throw an example in the docstring?

jdrew82: Done!

qduk: Sorry, I should've specified. Something like 

```python
Examples:
        >>> from netutils.ip import sort_list_ips
        >>> sort_list_ips("3.3.3.3,2.2.2.2,1.1.1.1")
        1.1.1.1,2.2.2.2,3.3.3.3
```

The docs pull it in so it needs to be in the same format as the others.

jdrew82: Oops, my bad! I should have noticed the others. Should be fixed now.

netutils/ip.py
```diff
@@ -596,3 +597,42 @@ def get_usable_range(ip_network: str) -> str:
...
+
+    Examples:
+        >>> from netutils.ip import sort_list_ips
+        >>> sort_list_ips("3.3.3.3,2.2.2.2,1.1.1.1")
+        '1.1.1.1/32,2.2.2.2/32,3.3.3.3/32'
```
itdependsnetworks, line 608:
```suggestion
        '1.1.1.1/32,2.2.2.2/32,3.3.3.3/32'
        >>> sort_list_ips("10.0.20.0/24,10.0.20.0/23,10.0.19.0/24")
        '10.0.19.0/24,10.0.20.0/23,10.0.20.0/24'
        >>> sort_list_ips("10.0.20.0/24,10.0.20.0/23,10.0.19.0/24", "interface")
        '10.0.19.0/24,10.0.20.0/23,10.0.20.0/24'
        >>> sort_list_ips("10.0.20.20/24,10.0.20.1/23,10.0.19.5/24", "interface")
        '10.0.19.5/24,10.0.20.1/23,10.0.20.20/24'
        >>> sort_list_ips(["10.0.20.20", "10.0.20.1", "10.0.19.5"], "address")
        '10.0.19.5,10.0.20.1,10.0.20.20'
```

itdependsnetworks: just looking to give more examples so people can understand how it is used.

netutils/ip.py
```diff
@@ -596,3 +597,42 @@ def get_usable_range(ip_network: str) -> str:
         lower_bound = str(net[1])
         upper_bound = str(net[-2])
     return f"{lower_bound} - {upper_bound}"
+
+
+def sort_list_ips(ips: t.Union[str, t.List[str]], sort_type: str = "network") -> str:
```
itdependsnetworks, line 602: I thought I hit send, but I think we should

* change the name, thinking either ip_sorted or ips_sorted
* This should return a list, imho.

jdrew82: Changing the name of the function to sort_ips makes sense IMO as that's what's being done, but returning a list of strings I don't think makes sense as the use cases I have for this are taking a list of strings, ie from a config context YAML and I'd want the string form of them for the generated template or for python where the same is being done. Returning a list of strings would require additional work in the template to concatenate the strings together.

netutils/ip.py
jdrew82: Got it, added these.

netutils/ip.py
itdependsnetworks: * the name follows the pattern in the rest of the file, i don't see why this would be different
* `Returning a list of strings would require additional work in the template to concatenate the strings together.` doing a `{{ (vals | ip_sorted).join(',') }}` seems like a reasonable trade off to keeping it a data structure vs an opinionated concatenation, as the next time you may want to do `{{ (vals | ip_sorted).join(';') }}` and would have to split and then rejoin.

jdrew82: I've renamed the function to `get_ips_sorted` to match the pattern of the other functions. I've also updated it to return a list instead of a concatenated string.

netutils/ip.py
```diff
@@ -596,3 +597,50 @@ def get_usable_range(ip_network: str) -> str:
         lower_bound = str(net[1])
         upper_bound = str(net[-2])
     return f"{lower_bound} - {upper_bound}"
+
+
+def get_ips_sorted(ips: t.Union[str, t.List[str]], sort_type: str = "network") -> t.List[str]:
+    """Given a concatenated list of CIDRs sorts them into the correct order and returns as concatenated string.
```
qduk, line 603:
```suggestion
    """Given a concatenated list of CIDRs sorts them into the correct order and returns them as a list.
```

qduk (approved):
LGTM!

itdependsnetworks (approved):

PR #483, opened 2024-03-14
Python 3.12 support

### LINKED ISSUE #280: Implement strtobool?
### Environment
* netutils version:  

### Proposed Functionality

Would look to reimplement the distutils strtobool function. 

### Use Case

The taking a string and getting a boolean from it is helpful. Could be used in YAML manipulation (yes/no), or for reading from environment settings. 

The distutils library that has the implementation is being retired and has deprecation warnings in it as of Python3.10. It would be nice to have a place to be able to get the same functionality from a trusted utility source. Understand if this is something that should not be here. Their advice is for each individual to implement their own.

Seeing this as an example:

```python
_MAP = {
    "y": True,
    "yes": True,
    "t": True,
    "true": True,
    "on": True,
    "1": True,
    "n": False,
    "no": False,
    "f": False,
    "false": False,
    "off": False,
    "0": False,
}


def strtobool(value):
    try:
        return _MAP[str(value).lower()]
    except KeyError:
        raise ValueError('"{}" is not a valid bool value'.format(value))
```

glennmatthews:
- ~Replace `distutils.version.LooseVersion` (`distutils` is removed in Python 3.12) with `looseversion.LooseVersion` (a very lightweight package that reimplements this specific class only)~
   - Instead, vendor `distutils.version` module into netutils.
- Replace `distutils.strtobool` usage in `tasks.py` with local code. (Relates to but does not close #280, as that was asking for a `strtobool` or `is_truthy` implementation in `netutils` itself)
- Require `flake8` 7.x as earlier versions didn't support Python 3.12
- Require `pylint` 3.x as earlier versions didn't support Python 3.12


pyproject.toml
```diff
@@ -28,6 +29,7 @@ include = [
 python = "^3.8"
 napalm = {version = "^4.0.0", optional = true}
 jsonschema = {version = "^4.17.3", optional = true}
+looseversion = "^1.3"
```
itdependsnetworks, line 32: I was planning on vendoring this, you can remove this one as we don't have any core requirements. If you can figure out a workaround for [crypt](https://docs.python.org/3/library/crypt.html) that would be great :)

glennmatthews: Not sure I understand what you're recommending. Should I vendor the code instead of adding the dependency as a part of this PR?

glennmatthews: Looks like an alternative approach to this PR would be to add `setuptools` as an explicit dependency, as setuptools vendors `distutils` into itself.

itdependsnetworks: This should work: https://github.com/networktocode/netutils/pull/484

glennmatthews: Thanks, rebased this branch under yours.

PR #489, opened 2024-03-19
Feature/platform mapper

bminnix:
Updating netutils for support of the nautobot-app-device-lifecycle-mgmt NIST URL creation for API 2.0

Updating the nist utility to only support certain OS PLatforms:
Cisco IOS
Cisco IOS XR
Cisco IOS XE
Cisco NXOS
Juniper Junos
Arista EOS


docs/dev/code_reference/nist.md
```diff
@@ -0,0 +1,17 @@
+# NIST URLs
+
+::: netutils.nist
+    options:
+        show_submodules: True
+
+**Notes:** 
+<br/>
+In order to use the URLs generated by netutils.nist.get_nist_urls*, you will need an api key provided by NIST [here]('https://nvd.nist.gov/developers/request-an-api-key').  This key will need to be passed in as an additional header in your request in the form of `{"apiKey": "<key_value>"}` as stated by NIST in their [Getting Started]('https://nvd.nist.gov/developers/start-here') section.
+<br/><br/>
```
itdependsnetworks, line 10: We should not need the breaks here.

docs/dev/code_reference/nist.md
```diff
@@ -0,0 +1,17 @@
...
+This module currently only supports the following OS platforms:
```
itdependsnetworks, line 11: Add a new line before and after here.

netutils/lib_mapper.py
```diff
@@ -545,6 +545,17 @@
     "paloalto_panos": "PAN_OS",
 }
 
+# Normalized | NIST Parser
```
itdependsnetworks, line 548: There should be a reverse in this file as well.

netutils/lib_mapper.py
```diff
@@ -545,6 +545,17 @@
...
+NIST_LIB_MAPPER_SUPPORTED = {
```
itdependsnetworks, line 549: Following standard, should just be `NIST_LIB_MAPPER`

mkdocs.yml
```diff
@@ -148,9 +148,11 @@ nav:
           - Library Helpers: "dev/code_reference/lib_helpers.md"
           - Library Mapping: "dev/code_reference/lib_mapping.md"
           - Mac Address: "dev/code_reference/mac.md"
+          - NIST: "dev/code_reference/nist.md"
           - OS Version: "dev/code_reference/os_version.md"
           - Password: "dev/code_reference/password.md"
           - Ping: "dev/code_reference/ping.md"
+          - Platform Mapper: "dev/code_reference/platform_mapper.md"
```
itdependsnetworks, line 155: This shouldn't be needed, see https://github.com/networktocode/netutils/pull/425/files#diff-3982ffdfe67538340b1567734143d0a1c02d63070279cc055d1e323a4301bcc7R37 for how we do this. You don't have to generate the md files themselves, just update that dictionary I referenced and then run `python development_scripts.py`.

bminnix: So, I'm a bit confused on this.

Running the pieces provided only created the lib_mapper maps for forward and reverse user docs right?

The documents in the `dev/code_reference` that were being added here were for the `nist.py `and `platform_mapper.py` modules themselves.

docs/dev/code_reference/nist.md
bminnix: I think this should also be resolved/outdated.  Requested changes made.

docs/dev/code_reference/nist.md
```diff
@@ -0,0 +1,18 @@
+# NIST URLs
+
+::: netutils.nist
+    options:
+        show_submodules: True
+
+**Notes:**
```
itdependsnetworks, line 7: thinking about it, this should be in the user guide. This will show up at the bottom of the code, I don't think many would look there.

bminnix: I've added some docs to the user guide that will hopefully work.

mkdocs.yml
itdependsnetworks: The name is confusing me, can we think of something other than platform mapper?

netutils/os_version.py
```diff
@@ -127,3 +128,93 @@ def compare_version_strict(current_version: str, comparison: str, target_version
...
+        parsed_version.get("build") is None or int(parsed_version.get("build")) <= 1  # type:ignore
+    ):
+        parsed_version["isfrs"] = True
+    elif parsed_version["type"].lower() == "r":
+        parsed_version["ismaintenance"] = True
+
+    return parsed_version
+
+
+os_version_parsers = {
```
itdependsnetworks, line 216: Let's move this to a function, that takes in a param for `os_version`.

itdependsnetworks: This would allow it to work as a Jinja Filter as well.

mkdocs.yml
bminnix: I'm not married to it, this was what it was being referred to up until it was made.  "If we had some kind of platform mapper that could..."

It's an abstract class builder for classes that need unified functionality.  I really don't know what to call that in regard to an actual module name.

Do you have any recommendations that would mesh well with the existing library?

netutils/os_version.py
bminnix: I've added a new function here and added to utils, does this meet what you had in mind?

netutils/os_version.py
```diff
@@ -127,3 +128,120 @@ def compare_version_strict(current_version: str, comparison: str, target_version
...
+    "juniper": {
+        "junos": juniper_junos_version_parser,
+    }
+}
+
+
+def vendorize_version(vendor: str, os_type: str, version: str) -> t.Dict[str, t.Any]:
```
itdependsnetworks, line 223: how about version_metadata?

netutils/os_version.py
```diff
@@ -127,3 +128,120 @@ def compare_version_strict(current_version: str, comparison: str, target_version
         >>>
     """
     return _compare_version(current_version, comparison, target_version, "strict")
+
+
+def juniper_junos_version_parser(version: str) -> t.Dict[str, t.Any]:
```
itdependsnetworks, line 133: how about juniper_junos_metadata?

itdependsnetworks:
@bminnix as we spoke about we will:

* Create an os_version_parser (or similar) and it will always provide major/minor/patch
* If you provide an OS, it will attempt to parse it futher with an OS specfic parser, like "if operating_system == "junos": return _junos_os_version_parser()
* keep all of the nist in itself

bminnix:
> @bminnix as we spoke about we will:
> 
> * Create an os_version_parser (or similar) and it will always provide major/minor/patch
> * If you provide an OS, it will attempt to parse it futher with an OS specfic parser, like "if operating_system == "junos": return _junos_os_version_parser()
> * keep all of the nist in itself

I believe all of the above is now factored in and ready for review.

bminnix:
> > @bminnix as we spoke about we will:
> > 
> > * Create an os_version_parser (or similar) and it will always provide major/minor/patch
> > * If you provide an OS, it will attempt to parse it futher with an OS specfic parser, like "if operating_system == "junos": return _junos_os_version_parser()
> > * keep all of the nist in itself
> 
> I believe all of the above is now factored in and ready for review.

@itdependsnetworks  Forgot to @ you

docs/dev/code_reference/platform_mapper.md
```diff
@@ -0,0 +1,5 @@
+# Platform Mapper
```
itdependsnetworks, line 1: remove

docs/user/lib_use_cases.md
```diff
@@ -21,10 +21,11 @@ Functions are grouped with like functions, such as IP or MAC address based funct
 - Library Helpers - Provides helpers to pull useful information, e.g. NAPALM getters.
 - Library Mapper - Provides mappings in expected vendor names between Netmiko, NAPALM, pyntc, ntc-templates, pyats, and scrapli.
 - MAC Address - Provides the ability to work with MAC addresses such as validating or converting to integer.
+- NIST - Provides the ability to obtain a URL formatted for NIST CPE Query.
 - OS Version - Provides the ability to work with OS version, such as defining an upgrade path.
 - Password - Provides the ability to compare and encrypt common password schemas such as type5 and type7 Cisco passwords.
 - Ping - Provides the ability to ping, currently only tcp ping.
-- Protocol Mapper - Provides a mapping for protocol names to numbers and vice versa.
+- Platform Mapper - Provides custom parsers for breakdown of OS Software Versions/Revisions.
```
itdependsnetworks, line 28: Shouldn't remove protocol, just add the platform mapper.

netutils/nist.py
```diff
@@ -0,0 +1,234 @@
...
+
+        # nist_urls.append(juniper:junos:12.2:r8-s2:*:*:*:*:*:*)
+        nist_urls.append(f"{base_ext}{_type}{_build}-{_service}{_service_build}{delim_seven}")
+
+        return nist_urls
+
+    raise ValueError("Failure creating Juniper JunOS Version. Format is unknown.")
+
+
+def get_nist_urls_default(os_platform_data: t.Dict[str, t.Any]) -> t.List[str]:
```
itdependsnetworks, line 159:
```suggestion
def get_nist_urls(os_platform_data: t.Dict[str, t.Any]) -> t.List[str]:
```

netutils/os_version.py
```diff
@@ -128,3 +129,192 @@ def compare_version_strict(current_version: str, comparison: str, target_version
...
+        parsed_version.get("build") is None or int(parsed_version.get("build")) <= 1  # type:ignore
+    ):
+        parsed_version["isfrs"] = True
+    elif parsed_version["type"].lower() == "r":
+        parsed_version["ismaintenance"] = True
+
+    return parsed_version
+
+
+def default_metadata(version: str) -> t.Dict[str, t.Any]:
```
itdependsnetworks, line 220: how about `version_metadata_basic`?

netutils/os_version.py
```diff
@@ -128,3 +129,192 @@ def compare_version_strict(current_version: str, comparison: str, target_version
         >>>
     """
     return _compare_version(current_version, comparison, target_version, "strict")
+
+
+def juniper_junos_metadata(version: str) -> t.Dict[str, t.Any]:
```
itdependsnetworks, line 134: version_metadata_juniper_junos

tests/unit/test_utils.py
```diff
@@ -24,7 +24,13 @@
 
 _EXCLUDED_DECORATOR_FUNCTIONS = ["wraps", "total_ordering", "abstractmethod"]
 
-_EXCLUDED_FUNCTIONS = ["jinja2_convenience_function", "import_module", "get_network_driver"]
+_EXCLUDED_FUNCTIONS = [
+    "jinja2_convenience_function",
+    "import_module",
+    "get_network_driver",
+    "get_nist_urls_default",
+    "get_nist_urls_juniper_junos",
```
itdependsnetworks, line 32: I think these can get removed.

docs/user/lib_use_cases_os_version.md
```diff
@@ -0,0 +1,40 @@
...
+
+- Default Parser
+- Juniper JunOS
+
+**See the following Default and Juniper JunOS parsed versions:**
+
+```python
+>>> from os_version import default_os_metadata, juniper_junos_metadata
+
+>>> default_os_metadata("cisco", "ios", "15.5")
```
itdependsnetworks, line 18: can update.

itdependsnetworks:
light cleanup... then i think we are good to merge!!

itdependsnetworks (approved):

jeffkala (approved):

Issue #492, opened 2024-03-20
Protocol Name to Number Map Missing Essentials


jdrew82:
### Environment
* Python version:  3.9
* netutils version:  1.7.0

### Expected Behavior
PROTO_NAME_TO_NUM list to include mappings for IPv4/v6, and ICMPv6.

### Observed Behavior
PROTO_NAME_TO_NUM list is missing IPv4, IPv6, and ICMPv6.

### Steps to Reproduce
1. Look at PROTO_NAME_TO_NUM list.
2.
3.

jeffkala:
fixed in #494

PR #493, opened 2024-03-20
Update Protocol Name to Number Map

jdrew82:
This PR is to address #492 and add in IPv4, IPv6, and ICMPv6 to the protocol name to number map.


itdependsnetworks:
We are missing like 20, was just easier to automate it moving forward with #494

jdrew82:
Works for me!

PR #519, opened 2024-05-06
Add Mappings for DNA Center

jdrew82:
This PR is to add the forward and reverse mappings for DNA Center. I'm a bit unsure how to handle IOS-XE for the reverse though as you can't have two `cisco_ios` keys so I changed the one for XE to be `cisco_xe`. I'd match it in the forward but the XE library mapping in Nautobot is missing some essential bits for IOS-XE.

### DIFF
--- a/development_scripts.py
+++ b/development_scripts.py
@@ -71,6 +71,18 @@ MAPPER = {
             "_dict": lib_mapper.CAPIRCA_LIB_MAPPER_REVERSE,
             "_file": "docs/user/lib_mapper/capirca_reverse.md",
         },
+        "dna_center": {
+            "header_src": "DNA_CENTER",
+            "header_dst": "NORMALIZED",
+            "_dict": lib_mapper.DNA_CENTER_LIB_MAPPER,
+            "_file": "docs/user/lib_mapper/dna_center.md",
+        },
+        "dna_center_reverse": {
+            "header_src": "NORMALIZED",
+            "header_dst": "DNA_CENTER",
+            "_dict": lib_mapper.DNA_CENTER_LIB_MAPPER_REVERSE,
+            "_file": "docs/user/lib_mapper/dna_center.md",
+        },
         "forwardnetworks": {
             "header_src": "FORWARDNETWORKS",
             "header_dst": "NORMALIZED",
--- a/development_scripts.py
+++ b/development_scripts.py
@@ -81,7 +81,7 @@ MAPPER = {
             "header_src": "NORMALIZED",
             "header_dst": "DNA_CENTER",
             "_dict": lib_mapper.DNA_CENTER_LIB_MAPPER_REVERSE,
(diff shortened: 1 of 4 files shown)


itdependsnetworks:
We generally do the XE in 1 direction, but not the other https://github.com/networktocode/netutils/blob/7e038fbc0b86a3a9fb410c6b9d5b2ed9e52c36f1/netutils/lib_mapper.py#L502

netutils/lib_mapper.py
```diff
@@ -124,6 +124,22 @@
...
 
+# DNA Center | Normalized
+DNA_CENTER_LIB_MAPPER = {
+    "IOS": "cisco_ios",
+    "IOS-XE": "cisco_ios",
+    "NX-OS": "cisco_nxos",
+    "IOS-XR": "cisco_xr",
+}
+
+# DNA Center | Normalized
```
itdependsnetworks, line 135: Can you flip this?

jdrew82: Oops, good catch! I meant to flip that. I've fixed it and remove the cisco_xe key.

itdependsnetworks:
Sorry, should have mentioned (and we should add to the docs), need to add to https://github.com/networktocode/netutils/blob/3b1004eaa66b00bf85475b3a8254c580e2a097e0/development_scripts.py#L37 and then run the development_script (if you run tests, it will tell you the command).

jdrew82:
> Sorry, should have mentioned (and we should add to the docs), need to add to https://github.com/networktocode/netutils/blob/3b1004eaa66b00bf85475b3a8254c580e2a097e0/development_scripts.py#L37 and then run the development_script (if you run tests, it will tell you the command). 

Doh! I always forget to do that for this project...

itdependsnetworks:
In fairness, we should have it documented.

development_scripts.py
```diff
@@ -71,6 +71,18 @@
...
+            "header_src": "DNA_CENTER",
+            "header_dst": "NORMALIZED",
+            "_dict": lib_mapper.DNA_CENTER_LIB_MAPPER,
+            "_file": "docs/user/lib_mapper/dna_center.md",
+        },
+        "dna_center_reverse": {
+            "header_src": "NORMALIZED",
+            "header_dst": "DNA_CENTER",
+            "_dict": lib_mapper.DNA_CENTER_LIB_MAPPER_REVERSE,
+            "_file": "docs/user/lib_mapper/dna_center.md",
```
itdependsnetworks, line 84:
```suggestion
            "_file": "docs/user/lib_mapper/dna_center_reverse.md",
```

I think this is the issue.

itdependsnetworks:
now you have to re-run development script.

jdrew82:
> now you have to re-run development script.

Sorry, meant to run that. Should be done now.

itdependsnetworks (approved):

PR #534, opened 2024-06-14
Extend ip_network extension to support method calls with kwargs

jeffkala:
ip_network has additional method calls that support additional kwargs to be passed in.  This adds that support.  For clarity, ip_address and ip_interface don't have these, so the extension is only on ip_network.

Example
```python
>>> list(ipaddress_network('192.168.1.0/28', 'subnets', new_prefix=30))
[IPv4Network('192.168.1.0/30'), IPv4Network('192.168.1.4/30'), IPv4Network('192.168.1.8/30'), IPv4Network('192.168.1.12/30')]
```

or jinja2

```python
from jinja2 import Environment, select_autoescape
from netutils.utils import jinja2_convenience_function

template_string = """
Example:
{% for sub in "192.168.1.0/28" | ipaddress_network('subnets', new_prefix=30) | list %}
{{ sub.with_netmask }}
{% endfor %}
"""

env = Environment(
    autoescape=select_autoescape()
)
env.filters.update(jinja2_convenience_function())
template = env.from_string(template_string)
result = template.render()
print(result)

```

```
>>> print(result)

Example:

192.168.1.0/255.255.255.252

192.168.1.4/255.255.255.252

192.168.1.8/255.255.255.252

192.168.1.12/255.255.255.252

```

### DIFF
--- a/netutils/ip.py
+++ b/netutils/ip.py
@@ -66,7 +66,7 @@ def ipaddress_network(ip: str, attr: str, **kwargs: t.Any) -> t.Any:
     Args:
         ip: IP network str compliant with ipaddress.ip_network inputs.
         attr: An attribute in string dotted format.
-        kwargs: If called a method allows for passing kwargs to resulting method call.
+        kwargs: Keyword arguments to pass along to the given method of ipaddress.ip_network.
 
     Returns:
         Returns the value provided by the ipaddress.ip_network attribute provided.
--- a/netutils/ip.py
+++ b/netutils/ip.py
@@ -2,7 +2,7 @@
 
 import ipaddress
 import typing as t
-from operator import attrgetter
+from operator import attrgetter, methodcaller
 
 from netutils.constants import IPV4_MASKS, IPV6_MASKS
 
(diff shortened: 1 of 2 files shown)


netutils/ip.py
```diff
@@ -60,12 +60,13 @@ def ipaddress_interface(ip: str, attr: str) -> t.Any:
...
 
 
-def ipaddress_network(ip: str, attr: str) -> t.Any:
+def ipaddress_network(ip: str, attr: str, **kwargs: t.Any) -> t.Any:
     """Convenience function primarily built to expose ipaddress.ip_network to Jinja.
 
     Args:
         ip: IP network str compliant with ipaddress.ip_network inputs.
         attr: An attribute in string dotted format.
+        kwargs: If called a method allows for passing kwargs to resulting method call.
```
itdependsnetworks, line 69:
```suggestion
        kwargs: Keyword arguments where the key is the method and the value is the value to send to the given method of ipaddress.ip_network.
```

jeffkala: so `attr` still remains the method.  kwargs are literally kwargs we're directly passing to the method that `attr` is specifying.

qduk (approved):
LGTM!

netutils/ip.py
jeffkala: you good leaving this one as is @itdependsnetworks ?

PR #540, opened 2024-06-26
Test bad parsing of duplicate line

itdependsnetworks:


itdependsnetworks:
To avoid any confusion, there is an additional commit, that is in PR #539 that should be merged there.

itdependsnetworks:
I will fix tests and get this PR updated and merged.

PR #569, opened 2024-08-27
failing test for banner stripping trailing newlines during parsing

jeffkala:


jeffkala:
Really this line is the problem https://github.com/networktocode/netutils/blob/ceb1693c7aaedcc9445a407bb0af7b26c0b16db6/netutils/config/parser.py#L226

Trailing newlines in a banner are valid in Cisco and in many cases like the login banner they're almost required to avoid the login banner from overrunning the `password` prompt.

Changing the line above to `rstrip(' ').` does fix the issue but causes all banner tests to fail so believe it would require a major release to make this change.

itdependsnetworks (approved):

PR #590, opened 2024-10-29
Adding support for Cisco ASA banner when nested underneath a parent config

huacchob:
Banner for Cisco ASA could be a child config, and would start with a white space. I added that support to both the BaseSpaceConfigParser.banner_start list and to CiscoConfigParser.regex_banner attribute

### DIFF
--- a/netutils/config/parser.py
+++ b/netutils/config/parser.py
@@ -53,7 +53,7 @@ class BaseSpaceConfigParser(BaseConfigParser):
     # pylint: disable=abstract-method
 
     comment_chars = ["!"]
-    banner_start = ["banner", "vacant-message"]
+    banner_start = ["banner", "vacant-message", " banner", "  banner"]
 
     def __init__(self, config: str):
         """Create ConfigParser Object.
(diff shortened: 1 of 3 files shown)


itdependsnetworks:
hey, can you provide test cases for this?

huacchob:
Would this test be enough?

netutils/config/parser.py
```diff
@@ -53,7 +53,7 @@ class BaseSpaceConfigParser(BaseConfigParser):
     # pylint: disable=abstract-method
 
     comment_chars = ["!"]
-    banner_start = ["banner", "vacant-message"]
+    banner_start = ["banner", "vacant-message", " banner", "  banner"]
```
itdependsnetworks, line 56: I don't understand this, why are we adding banner_start multiple times?

huacchob: The first one has has 1 white space and the second one has two.

itdependsnetworks: I would not think this is how we want to do this then. I would think we would understand the parent / child relationship would be how we determine it.

itdependsnetworks: Check in with @jmcgill298

jmcgill298: ya, that banner is a child of the group-policy, so it should be captured there. I guess for this case the banner is not multiline?

PR #601, opened 2024-12-03
Use legacycrypt where crypt isn't available

### LINKED ISSUE #594: Py 3.13 support?
### Environment
* netutils version: 1.10.0

Running tests on netutils in a python 3.13 enviroment indicates that it will no longer work: 

```
  /usr/lib/python3.13/site-packages/_pytest/assertion/rewrite.py:184: in exec_module
      exec(co, module.__dict__)
  tests/unit/test_password.py:5: in <module>
     from netutils import password
 netutils/password.py:4: in <module>
     import crypt  # pylint: disable=deprecated-module
 E   ModuleNotFoundError: No module named 'crypt'
```

### Proposed Functionality

tests should pass on python 3.13.

s-t-e-v-e-n-k:
Since Python 3.13 has removed the crypt module, add legacycrypt to the requirements if we're using that version of Python, importing it if the crypt module isn't found.

Fixes #594

### DIFF
--- a/netutils/password.py
+++ b/netutils/password.py
@@ -1,7 +1,5 @@
 """Functions for working with Passwords."""
 
-# TODO: Swap out crypt prior to py3.13
-import crypt  # pylint: disable=deprecated-module
 import random
 import secrets
 import string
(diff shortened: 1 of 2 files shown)


itdependsnetworks:
Thanks for this!!

just a few things

1. we have to make this optional. Can you add to `[tool.poetry.extras]` and make it `optional =true`
2. Change the import to only happen within the function and raise an error if not installed.

s-t-e-v-e-n-k:
> Thanks for this!!
> 
> just a few things
> 
> 1. we have to make this optional. Can you add to `[tool.poetry.extras]` and make it `optional =true`
> 2. Change the import to only happen within the function and raise an error if not installed.

Thanks for the super fast review, I've pushed up those changes.

itdependsnetworks:
Merged in #601

PR #602, opened 2024-12-16
Check only root level banner messages for ASA configs

huacchob:


huacchob:
I had a conversation with Jacob and he had agreed that this would be a solution. Would like some additional feedback please

itdependsnetworks:
Can you provide tests that show this working and to ensure we do not in the future regress on this?

jeffkala:
Internal chat about `banner_start` overload in ASA.  Otherwise this loks good to me

jeffkala (approved):

itdependsnetworks (approved):

Issue #158, opened 2022-10-21
abbreviated_interface_name support for Vlan interfaces


ryanmerolle:
### Environment
* netutils version:  1.2.0

### Proposed Functionality

Vlan interfaces are often commonly shortened.  It would be nice for the interface tool to do the same.

### Use Case

``` python
>>>print(abbreviated_interface_name("Vlan99"))
Vl99
```

ryanmerolle:
Interesting.  I was about to add this, but I see the constants are defined as https://github.com/networktocode/netutils/blob/develop/netutils/constants.py#L137-L141

```python
    "VLAN": "VLAN",
    "V": "VLAN",
    "Vl": "VLAN",
    "Vlan-interface": "VLAN",
    "vlan": "VLAN",
```

and https://github.com/networktocode/netutils/blob/develop/netutils/constants.py#L197

```python
    "VLAN": "Vl",
```

I would expect for _**interfaces**_ the mapping would be `"Vl": "Vlan"` and "Vlan": "Vl".  `VLAN` is mostly used for the VLAN definition and not the interface, but again I have been absorbed in Arista lately.

ryanmerolle:
Oddly enough when I do the following, I would expect an output of `Vl99`, but I see `VL99`:

```python
>>> vlan_interface = "Vlan99"
>>>print(abbreviated_interface_name(vlan_interface.upper()))
VL99
```

itdependsnetworks:
This works as well. 
```
>>> print(abbreviated_interface_name(canonical_interface_name("VLAN99")))
Vl99
>>>
```

Unfortunately, we need to anchor on something being correct. That being said, we have allowed you to override that definition of correct for this exact issue, see the signature

```
def abbreviated_interface_name(
    interface: str,
    addl_name_map: t.Optional[t.Dict[str, str]] = None,
    addl_reverse_map: t.Optional[t.Dict[str, str]] = None,
    verify: bool = False,
) -> str:
```

ryanmerolle:
This totally makes sense.  You have to use something as a basis to anchor.

To correct your example, I would need to adjust to this:

```python
>>> vlan_interface = "Vlan99"
>>> print(abbreviated_interface_name(canonical_interface_name(vlan_interface.upper())))
Vl99
```

PR #644, opened 2025-04-05
Add ability to get the lib mapper in a primary dictionary and a funct…

itdependsnetworks:
…ion to munge the data in a os centric view versus a lib centric view.

@glennmatthews this is largely driven and taken from: [here](https://github.com/nautobot/nautobot/blob/b80f037832b817407946ce96cd19d5d30df07db3/nautobot/dcim/constants.py#L103) and [here](https://github.com/nautobot/nautobot/blob/b80f037832b817407946ce96cd19d5d30df07db3/nautobot/dcim/utils.py#L100-L110). Want to make sure it will make sense as is to use within nautobot core in the future

@jdrew82 Note the update to dna_center. Will not remove anything until 2.x, but would be good to normalize this one in netutils or anywhere we are using it, such as ssot code.

### DIFF
--- a/netutils/lib_mapper.py
+++ b/netutils/lib_mapper.py
@@ -125,19 +125,25 @@ CAPIRCA_LIB_MAPPER_REVERSE: t.Dict[str, str] = {
 }
 
 # DNA Center | Normalized
-DNA_CENTER_LIB_MAPPER = {
+DNACENTER_LIB_MAPPER = {
     "IOS": "cisco_ios",
     "IOS-XE": "cisco_ios",
     "NX-OS": "cisco_nxos",
     "IOS-XR": "cisco_xr",
 }
 
+# REMOVE IN 2.X, kept for backward compatibility
+DNA_CENTER_LIB_MAPPER = copy.deepcopy(DNACENTER_LIB_MAPPER)
+
 # Normalized | DNA Center
-DNA_CENTER_LIB_MAPPER_REVERSE = {
+DNACENTER_LIB_MAPPER_REVERSE = {
     "cisco_ios": "IOS",
     "cisco_nxos": "NX-OS",
     "cisco_xr": "IOS-XR",
 }
+
+# REMOVE IN 2.X, kept for backward compatibility
+DNA_CENTER_LIB_MAPPER_REVERSE = copy.deepcopy(DNACENTER_LIB_MAPPER_REVERSE)
 
 # Normalized | Netmiko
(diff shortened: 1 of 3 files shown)


netutils/lib_mapper.py
```diff
@@ -641,3 +647,73 @@
...
+}
+
+
+def get_lib_mappings(
+    lib_mapping_reverse: t.Optional[t.Dict[str, t.Dict[str, str]]] = None
+) -> t.Dict[str, t.Dict[str, str]]:
+    """Generates a dictionary of all available network driver mappings.
+
+    Args:
+        lib_mapping_reverse: A reverse mapping of library names to their normalized names. Defaults to NAME_TO_LIB_MAPPER_REVERSE if not provided.
```
glennmatthews, line 693: What would be the context in which you'd provide a non-default `lib_mapping_reverse` value? I don't see any examples in this PR or in the tests.

netutils/lib_mapper.py
```diff
@@ -641,3 +647,73 @@
...
+
+    Returns:
+        A dictionary where the keys are normalized driver names and the values are dictionaries mapping tool names to their respective driver names.
```
glennmatthews, line 696: What's the value/purpose of having this as a function to call, rather than just having `LIB_MAPPINGS` (or whatever name) be another autogenerated constant?

tests/unit/test_lib_mapper.py
```diff
@@ -119,3 +130,47 @@ def test_lib_mapper_normalized_name(lib):
...
+    """Test for the lib_mapper definitions."""
+
+    def setUp(self):
+        """Gather all mapper dictionaries from the module."""
+        self.mappers = {}
+        self.reverse_mappers = {}
+        spec = importlib.util.spec_from_file_location("lib_mapper", lib_mapper.__file__)
+        lib_mapper_module = importlib.util.module_from_spec(spec)
+        sys.modules["lib_mapper"] = lib_mapper_module
+        spec.loader.exec_module(lib_mapper_module)
```
glennmatthews, line 145: What's the need for all this?

netutils/lib_mapper.py
itdependsnetworks: I was following a pattern that was similar to one described here: https://github.com/networktocode/netutils/issues/158#issuecomment-1287046776 where you could override certain variables, but let me think about it, perhaps it does not make sense.

glennmatthews (approved):

PR #648, opened 2025-04-23
add Dialer interface to interface type mappings

Uncle-Dee:


netutils/constants.py
```diff
@@ -15,6 +15,7 @@
     "Bd": "Bdi",
     "Bdi": "Bdi",
     "Bridge-Aggregation": "Port-channel",
+    "Di": "Dialer",
```
itdependsnetworks, line 18:
```suggestion
    "Dialer": "Dialer",
    "Di": "Dialer",
```

We always map the forward back to itself as well.

itdependsnetworks:
didn't mean to close it, just some git funkyness, handled in #678

PR #659, opened 2025-05-28
Add Nvidia Onyx config parser

apast0r:
This MR add a new parser for Nvidia/Mellanox onyx switches.

### DIFF
--- a/netutils/config/parser.py
+++ b/netutils/config/parser.py
@@ -1816,7 +1816,7 @@ class NvidiaOnyxConfigParser(BaseConfigParser):  # pylint: disable=abstract-meth
         return "\n".join(self._config_lines_only())
 
     def build_config_relationship(self) -> t.List[ConfigLine]:
-        r"""Parse text tree of config lines and their parents.
+        r"""Parse text config lines and banners.
 
         Examples:
             >>> from netutils.config.parser import BaseSpaceConfigParser, ConfigLine
@@ -1834,7 +1834,7 @@ class NvidiaOnyxConfigParser(BaseConfigParser):  # pylint: disable=abstract-meth
             ...        banner login "
             ...
             ...        MULTILINE BANNER"'''
-            >>> config_tree = BaseBraceConfigParser(config)
+            >>> config_tree = NvidiaOnyxConfigParser(config)
             >>> config_tree.build_config_relationship() == \
             ... [
             ...     ConfigLine(config_line='vlan 1', parents=()),


tests/unit/mock/config/parser/base/nvidia_onyx/onyx_full_sent.txt
```diff
@@ -0,0 +1,182 @@
...
+   snmp-server vrf default enable
+   
+##
+## Network management configuration
+##
+    stats sample telemetry enable
+# web proxy auth basic password ********
+   banner login "
+
+MULTI LINE BANNER"
```
itdependsnetworks, line 151: Can you explain this a bit?

apast0r: The configuration of onyx is simple. No indentation and all commands are single line.
Just one exception, the banner messages.
When you type a banner, onyx allows you to put between quotes any text that you wand, this includes new lines.
In the config the banner "`\nSOME THEXT\nMORE TEXT"` will appear as:

```
   banner login "
SOME TEXT
MORE TEXT"
```

This is quite abnormal, as you don't have a defined character for banner end.

Also, it automatically comments any line of config with sensitive content (such as passwords), this is why you see the: "# web proxy auth basic password **********". I haven't changed the config, it shows the password obfuscated.

apast0r: I was referring to the fact that the indentation does not affect the configuration.

You have this config:
```
##
## Network interface configuration
##
no interface mgmt0 dhcp
   interface mgmt0 ip address 192.168.0.2 /24
   interface mgmt1 shutdown
```
 Using the SpaceParser will not work, as it will try to add every command below the "no interface mgmt0 dhcp".
 mellanox just add 3 spaces to all lines that don't start with "no" to help the readability of the config.
 This is the same config:
```
no interface mgmt0 dhcp
interface mgmt0 ip address 192.168.0.2 /24
interface mgmt1 shutdown
```
 If banners doesn't exit (damn banners), I will just trim all the config and use the SpaceParser.
 
 About the parents of the banner, I do not understand the rules for considering one command a child of another.
 Given this banner:
```
    banner login "
SOME TEXT
MORE TEXT"
```
Why put:
```
ConfigLine(config_line="banner login "', parents=())
ConfigLine(config_line="SOME TEXT\MORE TEXT", parents=('banner login "'))
```
 
 And not:
```
ConfigLine(config_line='banner login "', parents=())
ConfigLine(config_line='SOME TEXT', parents=('banner login "'))
ConfigLine(config_line='MORE TEXT"', parents=('banner login "'))
```
or:
`ConfigLine(config_line='banner login "\nSOME TEXT\nMORE TEXT"', parents=())`

itdependsnetworks: The reason to do it with a single parent is, since this parser primary purpose is meant for comparing compliance between two different banners. With that in mind I can think of:

1. You can't replace partial of the banner in the config, so matching line 2 of expected with line 2 of intended doesn't matter if line 1 was not the same.
2. It made more sense to us (and still does) to make that a binary compliant or not

itdependsnetworks: Welcome to open another issue to change the behavior, but would require a major upgrade, and not something that would happen anytime soon. 

As such, please ensure this follows the same behavior as all the other parsers.

itdependsnetworks (approved):

PR #662, opened 2025-06-02
Nist updates

bminnix:
Fix: NXOS was not pulling CVEs because urls were using "nxos" but needed "nx-os"

Add: aruba_aos and paloalto_panos

Update: URLS were using parameter "virtualStringMatch", updated to use "cpeName" per the API documentation to get more accurate results


netutils/lib_mapper.py
```diff
@@ -609,7 +611,9 @@
 }
 
 # Normalized | NIST
+# TODO: This should be deprecated in next major release
```
itdependsnetworks, line 614: I think you can remove this, we would keep this regardless of `NIST_LIB_MAPPER_REVERSE_DICT`. This is the standard to map between all of the different OS's.

bminnix: Removed.

bminnix: I also updated the NIST_LIB_MAPPER_REVERSE to keep consistent entries with the others since the mapper itself is being kept for standard.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "netutils"
-version = "1.13.0"
+version = "1.14.0"
```
qduk, line 3: Can you remove this from your PR? We'll make a v1.14.0 prep PR where we'll bump the version and add any release notes since there will (most likely) be other updates.

netutils/lib_mapper.py
joaopsys, line 1: Are all the mappings in `NIST_TO_VENDOR` actually correct? Specially `os_name`

* `asa`: shouldn't it be `adaptive_security_appliance_software`
* `xe`: shouldn't it be `ios_xe`
* `xr`: shouldn't it be `ios_xr`

jeffkala (approved):

PR #665, opened 2025-06-17
Updata Interface Mappings

pato23arg:
Remove ambiguous mapping:

"B" -> "Bdi" -> causes conflicts, or interface misnaming for some platforms (e.g. HP Procurve Stacks)

Add missing Mappings (Bridge Virtual Interfaces - Cisco):

    "Bv": "BVI",
    "Bvi": "BVI",


netutils/constants.py
```diff
@@ -11,9 +11,10 @@
     "ap": "AppGigabitEthernet",
     "ATM": "ATM",
     "AT": "ATM",
-    "B": "Bdi",
```
itdependsnetworks, line 14: let's keep this for now, can revisit after, but this should get you moving.

netutils/constants.py
```diff
@@ -11,9 +11,10 @@
     "ap": "AppGigabitEthernet",
     "ATM": "ATM",
     "AT": "ATM",
```
jeffkala, line 13:
```suggestion
    "AT": "ATM",
    "B": "Bdi",
```

itdependsnetworks (approved):

PR #693, opened 2025-08-28
initial pass at rad etx config parser

### LINKED ISSUE #692: Add Rad ETX Parser
### Environment
* netutils version:  next

### Proposed Functionality
Support a parser for `rad_etx` driver.

### Use Case
1. Compliance and Parsing for Rad ETX platform

jeffkala:
Fixes #692


netutils/config/parser.py
```diff
@@ -1862,3 +1862,53 @@ def build_config_relationship(self) -> t.List[ConfigLine]:
...
+            >>> RadEtxConfigParser("config system virtual-switch").is_exit_or_exit_all("config system virtual-switch")
+            False
+            >>> RadEtxConfigParser("exit").is_exit_or_exit_all("exit")
+            True
+            >>>
+        """
+        for exit_command in ["exit", "exit all"]:
+            if line.lstrip().lower() == exit_command:
+                return True
+        return False
```
itdependsnetworks, line 1898:
```suggestion
        return line.lstrip().lower() in ["exit", "exit all"]
```
I think this should work

tests/unit/mock/config/parser/base/rad_etx/etx_full_received.py
```diff
@@ -0,0 +1,1262 @@
+from netutils.config.parser import ConfigLine
+
+data = [
+    ConfigLine(config_line='echo "Terminal Configuration"', parents=()),
```
itdependsnetworks, line 4: Should exit be returned?

tests/unit/mock/config/compliance/compliance/rad_etx/etx_basic_feature.py
```diff
@@ -0,0 +1,3 @@
+features = [
+    {"name": "terminal", "ordered": False, "section": ["terminal"]},
```
itdependsnetworks, line 2: Can you add a few more here? want to see how it works with nested results

tests/unit/mock/config/parser/base/rad_etx/etx_full_sent.txt
```diff
@@ -0,0 +1,732 @@
+    configure
+        echo "Terminal Configuration"
+    #       Terminal Configuration
```
itdependsnetworks, line 3: How are handling these? It's not really clear what the `#` signifies.

jeffkala: # is a comment character.  So basically these lines are getting removed before parsing.

netutils/config/parser.py
jeffkala: I had some oddities with this because it seems the rad device randomly returns some unfriendly newline chars.  Let me retest again though to make sure it wasn't a copy and paste issue during sanitization of the example config.

tests/unit/mock/config/parser/base/rad_etx/etx_full_received.py
jeffkala: my thought was no, that keeps the same pattern that we did in forinet_fortios which has a very similar structure, with exit/end statements.  We removed them in that case. Happy to reconsider it but not sure what the exit would provide.

PR #732, opened 2025-11-10
Adding initial LIBRENMS mappings for netutils

bminnix:
Adding a mapping for LibreNMS values to network_driver values for the initial values in use.


jeffkala:
Think you'll need to add a bit to the development_script to to make sure proper doc tables are auto generated since this is a new mapping.
https://github.com/networktocode/netutils/blob/b3e13d20fcf5a4e67b326734bccd0a6161c6a195/development_scripts.py#L37

netutils/lib_mapper.py
```diff
@@ -640,6 +650,16 @@
...
 }
 
+# Normalized | LibreNMS
+LIBRENMS_LIB_MAPPER = {
+    "arista_eos": "arista_eos",
+    "cisco_xe": "iosxe",
+    "cisco_nxos": "nxos",
+    "cisco_xr": "iosxr",
+    "juniper_junos": "junos",
+    "applogic_procera": "procera",
```
itdependsnetworks: Would this have ssh and be in netmiko? If so, let's try and reserve the name there, if not we need to add it to `_MAIN_LIB_MAPPER`

jeffkala (approved):
Looks like all the previous comments/issues were solved.

netutils/lib_mapper.py
jeffkala: I don't see a driver for applogic_procera in netmiko.

PR #751, opened 2025-12-18
Fix PAN-OS banner parsing by removing IOS-style banner end normalization

michalis1:
This PR fixes an issue in PaloAltoNetworksConfigParser where PAN-OS login banners were incorrectly normalized to use an IOS-style banner terminator (^C).
By removing banner end normalization, PAN-OS banners remain in their native quoted-string format and are parsed correctly.


yanshcodilime (approved):
LGTM

Issue #752, opened 2025-12-18
hp_comware config parsing with initial space


cheenpo:
### Environment
* netutils version:  1.15.1

### Expected Behavior
hp_comware config to be parsed

### Observed Behavior
`RENDER COMPLIANCE TASK GROUP failed: Validate the first line does not begin with a space `

### Steps to Reproduce
1. parse config from a hp_comware device where the first lines are:
```
#
 version 7.1.037, Release 2311P02
#
 sysname SW12345
#
```

PR #776, opened 2026-01-14
Added HP full config parsing

joewesch:
Closes: #752 

NTC-4804

This PR adds a more robust config example as well as some edge cases for the HP Comware platform and adds a custom parser method to properly parse the different sections.

I also was confused which class to update initially, but then I noticed that the HPComwareConfigParser class wasn't really doing much by itself and was redundant. I consolidated the two classes and then added some deprecation warnings and updated the parser to directly use the HPEConfigParser class.

### DIFF
--- a/netutils/config/compliance.py
+++ b/netutils/config/compliance.py
@@ -35,6 +35,12 @@ parser_map: t.Dict[str, t.Type[parser.BaseConfigParser]] = {
 }
 
 
+# Network OSes that we do not strip leading whitespace from the config lines.
+NON_STRIP_NETWORK_OS = [
+    "hp_comware",
+]
+
+
 # TODO: Once support for 3.7 is dropped, there should be a typing.TypedDict for this which should then also be used
 # as the return type for a bunch of the following methods.
 default_feature: t.Dict[str, t.Union[str, bool, None]] = {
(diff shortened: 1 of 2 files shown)


itdependsnetworks:
> > can you add a compliance check here: https://github.com/networktocode/netutils/tree/develop/tests/unit/mock/config/compliance/compliance
> > I am interested to see how you define the scope of the configs when there is a space at the start of the line, and would ask that we consider how that will work with GC
> 
> Hmm, ok. In that case I'd probably just `lstrip()` all config lines without parents.

That **may** come with other issues for compliance or remediation. Once you put in a compliance test I will take a look at how it works. Pinging @jeffkala on this as well.

tests/unit/mock/config/compliance/compliance/hp_comware/hp_comware_full_backup.txt
```diff
@@ -0,0 +1,280 @@
...
+ ntp-service unicast-server 192.0.2.251 vpn-instance internal
+ ntp-service unicast-server 192.0.2.182 vpn-instance internal
+#
+acl number 1234 name NTC-ACL
+ rule 0 permit tcp destination 0.0.0.0 255.255.255.0 destination-port range 1000 2000
+ rule 0 comment hello world
+#
+ tftp client source interface M-GigabitEthernet0/0/0
+#
+ header motd %
```
jeffkala: Is header actually indented 1 space? Seems odd that its nested under seemingly the acl parent?

PR #798, opened 2026-01-29
update lib mappers for 2026 reconciliation

jeffkala:


netutils/lib_mapper.py
```diff
@@ -495,6 +536,19 @@
     "SRX": "juniper_junos",  # no reverse
 }
 
+# NCCLIENT | Normalized
+NCCLIENT_LIB_MAPPER: t.Dict[str, str] = {
```
joewesch, line 540: You need to add this here: 
https://github.com/networktocode/netutils/blob/fd68af083a91db198def042e20eeaf8f02087fed/tests/unit/test_lib_mapper.py#L8-L22

netutils/lib_mapper.py
```diff
@@ -694,6 +751,23 @@
...
+    "cisco_nxos": "nexus",
+    "cisco_xe": "iosxe",
+    "cisco_xr": "iosxr",
+    "h3c_comware": "h3c",
+    "huawei": "huawei",
+    "huawei_vrp": "huaweiyang",
+    "huawei_vrpv8": "huaweiyang",
+    "hp_comware": "hpcomware",
+    "juniper_junos": "junos",
+    "nokia_sros": "alu",
```
joewesch, line 768: Copy pasta?

docs/user/lib_mapper/ncclient.md
```diff
@@ -0,0 +1,11 @@
+| NCCLIENT | | NORMALIZED |
```
joewesch, line 1: Do we need to add anything to `docs/user/lib_use_cases_lib_mapper.md`?

jeffkala: yes, thanks.

netutils/lib_mapper.py
jeffkala: done

netutils/lib_mapper.py
jeffkala: which part? This one is accurate from what I've looked at. Nokia_sros doesn't have a dedicated driver, it uses the `alu` original (alcatel_sros) driver.

joewesch: Ah, sorry I may be wrong. I saw that `NAPALM_LIB_MAPPER_REVERSE` and `PYATS_LIB_MAPPER_REVERSE` both have `"nokia_sros": "sros"`.

joewesch (approved):

Issue #803, opened 2026-02-03
Palo Alto parser raise exception for empty configurations


joewesch:
### Environment
* Python version:  3.11.13
* netutils version:  1.17.1a0

### Expected Behavior
I expect to be able to load an empty configuration into the Palo Alto Parser and it not raise an error to match other parsers (e.g., IOSConfigParser)

### Observed Behavior
The parser raises an exception that the configuration is not in the proper format (a blank config has no format)

### Steps to Reproduce
```python
In [1]: from netutils.config.parser import IOSConfigParser, PaloAltoNetworksConfigParser

In [2]: IOSConfigParser("")
Out[2]: <netutils.config.parser.IOSConfigParser at 0x7ba6b3eabb50>

In [3]: PaloAltoNetworksConfigParser("")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 PaloAltoNetworksConfigParser("")

File ~/netutils/netutils/config/parser.py:66, in BaseSpaceConfigParser.__init__(self, config)
     60 """Create ConfigParser Object.
     61
     62 Args:
     63     config (str): The config text to parse.
     64 """
     65 self._indent_level = 0
---> 66 super(BaseSpaceConfigParser, self).__init__(config)

File ~/netutils/netutils/config/parser.py:33, in BaseConfigParser.__init__(self, config)
     31 self.generator_config = (line for line in self.config_lines_only.splitlines())
     32 self.config_lines: t.List[ConfigLine] = []
---> 33 self.build_config_relationship()

File ~/netutils/netutils/config/parser.py:1567, in PaloAltoNetworksConfigParser.build_config_relationship(self)
   1565     self.generator_config = (line for line in list_config)
   1566 elif not any(line.startswith("set ") for line in config_lines):
-> 1567     raise ValueError("Unexpected configuration format. Please provide in 'set' or 'default' (brace) format.")
   1569 # build config relationships
   1570 for line in self.generator_config:

ValueError: Unexpected configuration format. Please provide in 'set' or 'default' (brace) format.
```

PR #804, opened 2026-02-03
Fixed PanOS empty config parsing

joewesch:
Closes: #803 

The `PaloAltoNetworksConfigParser` was written contradictory to the rest of the parsers in that it raised an error when the configuration was empty. This PR changes that behavior to bring it in line with the rest of the parsers. I also added a unit test to verify that all parsers allow empty configurations and return an empty list.

NTC-4957

### DIFF
--- a/tests/unit/test_parser.py
+++ b/tests/unit/test_parser.py
@@ -100,7 +100,7 @@ def test_leading_spaces_config_start(network_os):
 
 @pytest.mark.parametrize("network_os", all_network_os)
 def test_empty_config(network_os):
-    "Test that an empty config returns an empty list and does not raise an error."
+    """Test that an empty config returns an empty list and does not raise an error."""
     config = ""
     os_parser = compliance.parser_map[network_os]
     assert os_parser(config).config_lines == []


tests/unit/test_parser.py
```diff
@@ -13,17 +13,18 @@
...
-    for _file in glob.glob(f"{MOCK_DIR}/find_children_w_parents/{network_os}/*{TXT_FILE}"):
-        find_children_w_parents_parameters.append([_file, network_os])
+all_network_os = list(compliance.parser_map.keys())
+for _network_os in all_network_os:
+    for _file in glob.glob(f"{MOCK_DIR}/base/{_network_os}/*{TXT_FILE}"):
+        base_parameters.append([_file, _network_os])
+    for _file in glob.glob(f"{MOCK_DIR}/find_all_children/{_network_os}/*{TXT_FILE}"):
+        find_all_children_parameters.append([_file, _network_os])
+    for _file in glob.glob(f"{MOCK_DIR}/find_children_w_parents/{_network_os}/*{TXT_FILE}"):
+        find_children_w_parents_parameters.append([_file, _network_os])
```
joewesch: This is just changing `network_os` to `_network_os` as the loop variable so that we don't perpetuate copy/pasting the same pylint disable comment over and over...

jeffkala (approved):

PR #812, opened 2026-02-11
Provides a system to determine available interfaces in a range

abates:
This request adds a method to slice lists of interfaces in order to facilitate provisioning. Sometimes upstream connections are placed within a given range of interfaces. For instance, a primary connection may always go in the range of Ethernet[1-24] and a secondary connection may go into Ethernet[25-48]. If a switch has all the interfaces Ethernet[1-48] available, then the two needed slices are returned. If a subset of needed interfaces are available, then only those interfaces in the subset are returned. For instance:

```python
slice_interface_range(["Ethernet2", "Ethernet3", "Ethernet4", "Ethernet6", "Ethernet7", "Ethernet8"], "Ethernet1", "Ethernet5")
```

will effectively produce:
```python
["Ethernet2", "Ethernet3", "Ethernet4"], ["Ethernet6", "Ethernet7", "Ethernet8"]
```

### DIFF
--- a/netutils/interface.py
+++ b/netutils/interface.py
@@ -458,8 +458,7 @@ def sort_interface_list(interfaces: t.List[str]) -> t.List[str]:
     return list(_iter_tree(root, []))
 
 
-def slice_interface_range(
-    interface_names: t.List[str], *cut_points: str) -> t.Tuple[t.Iterator[str], ...]:
+def slice_interface_range(interface_names: t.List[str], *cut_points: str) -> t.Tuple[t.Iterator[str], ...]:
     """Slice a sorted list of interface names based on cut points.
 
     The interface names will be cut based on the values in the cut_points list. Each cut
@@ -519,11 +518,13 @@ def slice_interface_range(
                 yield None
 
     interface_names = sort_interface_list(interface_names)
-    sorted_cut_points = sort_interface_list(cut_points) # type: ignore
+    sorted_cut_points = sort_interface_list(cut_points)  # type: ignore
     slices = []
     start_index = None
     for i, interface in enumerate(interface_names):
-        if interface == sorted_cut_points[0] or _split_interface_tuple(sorted_cut_points[0]) < _split_interface_tuple(interface):
+        if interface == sorted_cut_points[0] or _split_interface_tuple(sorted_cut_points[0]) < _split_interface_tuple(
+            interface
+        ):
             if start_index is not None:
                 slices.append(interface_name_yielder(interface_names[start_index:i]))
             start_index = i
--- a/netutils/interface.py
+++ b/netutils/interface.py
@@ -458,6 +458,86 @@ def sort_interface_list(interfaces: t.List[str]) -> t.List[str]:
     return list(_iter_tree(root, []))
 
 
+def slice_interface_range(
+    interface_names: t.List[str], *cut_points: str) -> t.Tuple[t.Iterator[str], ...]:
+    """Slice a sorted list of interface names based on cut points.
+
+    The interface names will be cut based on the values in the cut_points list. Each cut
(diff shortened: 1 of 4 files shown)


netutils/interface.py
```diff
@@ -458,6 +458,86 @@ def sort_interface_list(interfaces: t.List[str]) -> t.List[str]:
...
 
 
+def slice_interface_range(
+    interface_names: t.List[str], *cut_points: str) -> t.Tuple[t.Iterator[str], ...]:
+    """Slice a sorted list of interface names based on cut points.
+
+    The interface names will be cut based on the values in the cut_points list. Each cut
+    point defines the start of its respective slice.  A slice runs from its cut point
+    up to (but not including) the next cut point.  The final slice contains all
+    remaining interface names from the last cut point to the end of the sorted list. This
```
itdependsnetworks, line 467: This makes sense to me, but in my mind should also be extended to start. Having to always define the first element (Ethernet1 in tests below) seems off to me.

I could understand either both start and end needing to be explicit or neither being explicit, but start being explicit and end being implicit is a bit confusing to me.

mpenning:
Suggestions for an API enhancement... using [`ciscoconfparse2`][1] for API reference implementation...

Full-disclosure, I'm the [`ciscoconfparse2`][1] owner...

Don't fear the GPL in my case... if I was going to fuss about GPL viral licensing, Cisco would already be hot sauce for using it all over the place with non-GPL licensing.

```python
❯ python
Python 3.13.0 (main, Dec  1 2024, 17:40:20) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
>>> from ciscoconfparse2 import CiscoRange
>>>
>>> intfs = CiscoRange("Ethernet2-4,6-8")
>>> for intf in intfs:
...     print(intf)
...
Ethernet2
Ethernet3
Ethernet4
Ethernet6
Ethernet7
Ethernet8
>>>
>>> CiscoRange("Ethernet2-4") in intfs
True
>>>
>>> CiscoRange("Ethernet47-48") in intfs
False
>>>
```

  [1]: https://github.com/mpenning/ciscoconfparse2

itdependsnetworks:
@mpenning wanted to start with a quick thanks for your open source contributions! Appreciate and notice all of your hard work over the years

> Don't fear the GPL in my case

Our lawyers say otherwise, so while I understand the spirit of what you are saying, in practice my hands our tied.

mpenning:
@itdependsnetworks 

The point was twofold... 

- Yes, just use the GPL `CiscoRange()` if there were no such things as lawyers
- If you don't adopt the GPL code, you can still use the `CiscoRange()` API, which doesn't require manually touching lists of string interface names

netutils/interface.py
abates: Alright, I think my intent maybe wasn't very clear. 

The idea is that, for example, I always connect access ports (servers) to switch ports `Ethernet1-Ethernet24`, and so the `slice_interface_range` would look for all of the interfaces in that range that are in the available `interface_names` list. The starting point doesn't have to be at the beginning, though, it could be a range of `Ethernet16-Ethernet32`.

The list of available interfaces may be a superset of the ranges provided in the cut list, so I can't really think of a way to implicitly choose a start and end. For instance, if the available interfaces list is `Ethernet1-Ethernet48` and my intended range is `Ethernet16-Ethernet32` how would we run this function with either of those values implied?

I think what you are suggesting is that this should work similar to python list slicing, which allows no start and/or end allowing implied start and end of `0` and `-1`. This is a language syntax convention, and I don't know how to extend that concept to function arguments.

I'm certainly open to changing this behavior. Can you provide some examples of how you'd like to see the method used with implied start and/or ends?

PR #837, opened 2026-05-27
Updating the find_children_w_parents method and tests

bminnix:
Fixed `BaseSpaceConfigParser.find_children_w_parents()` to no longer include sibling lines that do not match the child pattern.


jeffkala:
Seems like the the concept makes sense, but all the test for `*_received.txt` files are in theory invalid config. Shouldn't the received still have the proper config structure?

PR #862, opened 2026-07-20
adding fix to skip the update of arg keys/values when an error is pre…

huacchob:
…sent and type annotation updates

### DIFF
--- a/netutils/nist.py
+++ b/netutils/nist.py
@@ -9,7 +9,7 @@ from netutils.lib_mapper import NIST_LIB_MAPPER_REVERSE
 from netutils.os_version import version_metadata
 
 # Setting up the dataclass values for specific parsers
-PLATFORM_FIELDS: t.Dict[str, t.Any] = {
+PLATFORM_FIELDS: dict[str, t.Any] = {
     "default": [
         ("vendor", str),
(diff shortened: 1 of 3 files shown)


netutils/nist.py
```diff
@@ -38,15 +38,15 @@
 }
 
 
-class OsPlatform(metaclass=abc.ABCMeta):
+class OsPlatform(abc.ABC):
```
jeffkala, line 41: so largely this is the only actual change? All the mypy typing changes are not needed? Just updates based on newer floor of python version?

huacchob: The change I actually want to implement is in netutils.nist.py line 221 https://github.com/networktocode/netutils/pull/862/changes#diff-e90ca65a37a941e8f2767f35f4b6483dbccc5fb9ca85d74dcc904bb28ac7a748R221

This skips adding the "Error" key to the field_values dict.

This is the error I get when the "Error" key is present:

```
[12:43:55.465] There was an error in creating URLs for this Software Version. Please check the version value for Cisco cisco_ios 03.03.05SE. ERROR: CiscoIos.__init__() got an unexpected keyword argument 'Error'
```

PR #890, opened 2026-08-21
Fix NIST platform builder for unparseable versions

### LINKED ISSUE #861: Error key from returned dict incorrectly passed as argument to dynamically created dataclass
### Environment
* Python version:  3.11
* netutils version:  1.17.1

### Expected Behavior
When I run the netutils.nist._os_platform_object_builder function, I expect this to properly call the dynamically created dataclass.

### Observed Behavior
If the netutils.os_version.version_metadata function is called, and the version can't be properly parsed, a dictionary is returned instead, which contains the `Error` key. This would then get added to the dynamically created dataclass, and we get an unexpected keyword argument error, a TypeError.

Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/opt/nautobot/lib64/python3.11/site-packages/netutils/nist.py", line 231, in _os_platform_object_builder
    return platform_cls(**field_values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: CiscoIos.__init__() got an unexpected keyword argument 'Error'
>>>

NataliaPerez08:
Fixes #861

`version_metadata()` may return a dictionary containing keys that are not
defined on the dynamically generated platform dataclass when a version cannot
be parsed.

This change filters the returned metadata to fields defined by the platform
dataclass before constructing the object.

A regression test was added for the Cisco IOS version `03.03.05SE`.

Tests:
- 809 passed
- 10 skipped

### DIFF
--- a/netutils/os_version.py
+++ b/netutils/os_version.py
@@ -258,9 +258,9 @@ def _basic_version_metadata(version: str) -> t.Dict[str, t.Any]:
     basic_regex: re.Pattern[str] = re.compile(
         r"""
         ^
-        (?P<major>0|[1-9]\d*)
+        0*(?P<major>\d+)
         \.
-        (?P<minor>0|[1-9]\d*)?
+        (?:0*(?P<minor>\d+))?
         .*$
         """,
         re.VERBOSE,
--- a/tests/unit/test_os_versions.py
+++ b/tests/unit/test_os_versions.py
@@ -50,6 +50,30 @@ PLATFORM_VERSION_METADATA = [
             "vendor_metadata": False,
         },
     },
+    {
+        "sent": {"vendor": "cisco", "platform": "ios", "version": "03.03.05SE"},
+        "received": {
+            "major": "3",
+            "minor": "3",
+            "vendor_metadata": False,
+        },
+    },
+    {
+        "sent": {"vendor": "cisco", "platform": "ios", "version": "3.03.05SE"},
+        "received": {
+            "major": "3",
+            "minor": "3",
+            "vendor_metadata": False,
+        },
+    },
+    {
+        "sent": {"vendor": "cisco", "platform": "ios", "version": "3.3.05SE"},
+        "received": {
+            "major": "3",
+            "minor": "3",
+            "vendor_metadata": False,
+        },
+    },
     # Juniper Junos uses a custom parser
     {
         "sent": {"vendor": "juniper", "platform": "junos", "version": "12.4R"},


jeffkala:
@NataliaPerez08 thanks for the PR! Quick question. Is the version in the regression test actually coming out of `show version` or similar from an actual legacy device?

Just curious as the leading zero in the major version field is actually not following semver: https://semver.org/spec/v2.0.0.html#spec-item-2

Trying to ensure that this isn't a "made up failure" vs an actual bug.

NataliaPerez08:
Thanks for raising that. I checked the provenance of the test value.

03.03.05SE is a real Cisco IOS XE release, not a synthetic SemVer value. Cisco documents it for Catalyst 3650/3850 and maps it to a separate IOSd version.

That said, you are right that the test currently passes it with platform="ios", while netutils distinguishes IOS XE in some of its internal library mappings (cisco_xe vs cisco_ios).

So the version itself is valid, but I want to make sure the platform value in this regression test reflects the intended NIST API semantics. I can update the test to use the appropriate IOS XE platform identifier if that is preferred.

jeffkala:
guess my point here is if cisco see 03.03.05SE as valid. Netutils should be able to parse it successfully. 

Meaning update the basic_regex to be able to find leading zeros and then strip them.  Like this

```python
    # If version is not SemVer 2.0.0, attempt to find major/minor only.
    basic_regex: re.Pattern[str] = re.compile(
        r"""
        ^
        0*(?P<major>\d+)
        \.
        (?:0*(?P<minor>\d+))?
        .*$
        """,
        re.VERBOSE,
    )
    ```
    
    Which new regex fixes it to
    ```
>>> from netutils.nist import version_metadata
>>> version_metadata("cisco", "ios", "03.03.05SE")
{'major': '3', 'minor': '3', 'vendor_metadata': False}
>>> 
>>> version_metadata("cisco", "ios", "3.03.05SE")
{'major': '3', 'minor': '3', 'vendor_metadata': False}
>>> version_metadata("cisco", "ios", "3.3.05SE")
{'major': '3', 'minor': '3', 'vendor_metadata': False}
```

NataliaPerez08 (commented):
Thanks for the feedback. I reworked the fix so the parsing issue is handled in `version_metadata()` rather than downstream in the NIST builder.

The fallback regex now accepts and normalizes leading zeros, and I added regression coverage for `03.03.05SE`, `3.03.05SE`, and `3.3.05SE`.

The full test suite passes: 828 passed, 2 skipped.

PR #900, opened 2026-09-02
Fix find_children_w_parents to include only real descendants

### LINKED ISSUE #836: Bug: BaseSpaceConfigParser.find_children_w_parents()
### Environment
* Python version:  3.12.4
* netutils version:  1.17.2a0

### Expected Behavior
Defining the parent_pattern and child_pattern returns only child lines that are in the parent even if both parent and child are matched.
```
>>> parsed_object.find_children_w_parents(parent_pattern="context local", child_pattern="  logging", match_type="regex")
['  logging something', '  logging something else']
```

### Observed Behavior
Defining the parent_pattern and child_pattern returns everything under the parent.
```
parsed_config.find_children_w_parents(parent_pattern="context local", child_pattern="  logging", match_type="regex")
['  something not logging', '  logging something', '  logging something else', '  something else not logging']
```

### Steps to Reproduce
``` 
>>> from netutils.config.parser import BaseSpaceConfigParser

>>> config = '''context local
 something not logging
 logging something
 logging something else
 something else not logging
'''

>>> parsed_object = BaseSpaceConfigParser(config)
>>> parsed_config.find_children_w_parents(parent_pattern="context local", child_pattern=" logging", match_type="regex")

### PR DESCRIPTION
**Pull request by kat-git-hub**
Fixes #836 

find_children_w_parents grouped lines by their top level parent instead of the actual matched line. So matching one child pulled in every sibling under the same parent, even ones that don't match the pattern at all.

Example from the issue:
```python
parsed.find_children_w_parents(parent_pattern="context local", child_pattern=" logging", match_type="regex")
```
Before it also returned "something not logging". Now only the two logging lines come back.

Also updated the existing ios_full_received.txt fixture. It expected contact-email-addr to be included alongside profile "CiscoTAC-1", but contact-email-addr doesn't match the test's child_pattern at all, it's just a sibling under call-home. Same bug, just baked into the expected test output. Added a separate flat_siblings test case for the exact issue scenario.

### DIFF
--- a/changes/836.fixed.md
+++ b/changes/836.fixed.md
@@ -1,0 +1,1 @@
+Fixed `find_children_w_parents` to only return the matched child line and its actual descendants, instead of every sibling line under the same top-level parent.
--- a/netutils/config/parser.py
+++ b/netutils/config/parser.py
@@ -416,20 +416,18 @@ class BaseSpaceConfigParser(BaseConfigParser):
             >>> print(bgp_conf)
             ['  address-family ipv4 unicast', '   neighbor 192.168.1.2 activate', '   network 172.17.1.0 mask']
         """
-        config = []
-        potential_parents = [
-            elem.parents[0]
-            for elem in self.build_config_relationship()
+        relationships = self.build_config_relationship()
+        matched_lines = {
+            elem.config_line
+            for elem in relationships
             if self._match_type_check(elem.config_line, child_pattern, match_type)
-        ]
-        for cfg_line in self.build_config_relationship():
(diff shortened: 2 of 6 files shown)


jeffkala:
Thanks for the contribution @kat-git-hub; how is this PR different than https://github.com/networktocode/netutils/pull/837

kat-git-hub:
@jeffkala  Sorry, I hadn't seen #837 before opening this PR. 
From what I can tell, #837 started out fixing the same bug, but a later commit added the parent config line to the output too, seemingly to support a Golden Config use case. This PR sticks to exactly what the issue asked for, only the matched child lines and their descendants, without the parent line.

PR #914, opened 2026-09-22
Fixed `sanitize_config_jinja` to use an explicit opt-in boolean

joewesch:
This is a follow up to fix a breaking change bug introduced in #846 and https://github.com/nautobot/nornir-nautobot/pull/298. I inadvertently triggered the the `sanitize_config_jinja` function when users in Golden Config were replacing their config with post-processing jinja templates.

The solution, to be both backwards compatible and add the feature so they can co-exist, is to change the "jinja awareness" to be explicitly opt-in instead of automatic. This PR adds a new key/value pair that is expected to be a part of the filters list rather than solely relying on the presence of `{{`. This makes it so they can opt-in to the feature while still using `{% raw %}...{% endraw %}` tags to surround any post-processing they want to use together with other jinja filters.

Once this is merged, I will make a similar change in `nornir-nautobot` to remove the automatic inclusion as well. Luckily, this has not been fully adopted or forced out to Golden Config users yet and they only get this breaking change if they had updated the `nornir-nautobot` dependency to v4.4.0+.


netutils/config/clean.py
```diff
@@ -92,27 +104,61 @@ def sanitize_config(config: str, filters: t.Optional[t.List[t.Dict[str, str]]] =
...
-    (`pip install netutils[optionals]`).
+    (`pip install netutils[optionals]`) when at least one filter opts in.
 
     Args:
         config: A string representation of a device configuration.
-        filters: A list of dictionaries of regex patterns and Jinja-aware replacement
-            templates used to sanitize configuration. Defaults to an empty list.
+        filters: A list of dictionaries of regex patterns and replacement templates used to
+            sanitize configuration, each optionally setting `jinja` to `True` to render its
+            replacement as a Jinja template. Defaults to an empty list.
```
jeffkala, line 161: I know its a carry over but the `Default to an empty list` actually seems wrong. Doesn't it just default to `None`?  Looking at the code it appears to.

netutils/config/clean.py
```diff
@@ -122,8 +168,9 @@ def sanitize_config_jinja(config: str, filters: t.Optional[t.List[t.Dict[str, st
         >>> config = "username admin privilege 15 secret 9 SuperSecret"
         >>> SANITIZE_FILTERS = [
         ...     {
-        ...         "regex": r"^username (\S+) privilege 15 secret 9 (\S+)$",
-        ...         "replace": r"username {{ \1 }} privilege 15 secret 9 {{ \2 | hash_data('md5') }}",
+        ...         "regex": r"^(username \S+ privilege 15 secret 9 )(\S+)$",
+        ...         "replace": r"\1{{ \2 | hash_data('md5') }}",
+        ...         "jinja": True,
```
jeffkala, line 173: Any chance this introduces a namespace issue somewhere in the library path?  I don't think it would but trying to be sure.  Maybe we do `render_jinja` or something instead?

netutils/config/clean.py
joewesch: Yup, changed. Thanks.

netutils/config/clean.py
joewesch: Makes sense, done.

jeffkala (approved):

